Issue #33, opened 2019-04-04
New command: wit resolve


samh-sifive:
dry-run of wit update -- may need to `git fetch` if requested commit is not present in git repo.

aaronjanse:
Do we need this anymore now that we have `wit inspect`?

PR #72, opened 2019-06-12
implement threaded dependency tree updating

aaronjanse:
Even though it passes all the tests, this is a big diff so we need to be careful deciding whether to merge this, of course.

### DIFF
--- a/lib/gitrepo.py
+++ b/lib/gitrepo.py
@@ -4,6 +4,7 @@ import subprocess
 from pathlib import Path
 import logging
 from pprint import pformat
+import datetime
 import json
 import sys
 import lib.manifest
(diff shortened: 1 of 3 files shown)


aaronjanse:
Let's say we have the following dependency tree:
```
├─ bar
│  └─ abc:1.9.2
│     └─ xyz
└─ foo
   └─ abc:2.1.4
```

Threaded cloning would see `abc:1.9.2`, clone `xyz`, then replace `abc:1.9.2` with `abc:2.1.4`, _but it wouldn't delete `xyz`_.

In order to fix this, I think we'd need to keep track of a dependency tree of what's been cloned so far.
While tedious, we'd need this for #76 and #33 anyway.

jackkoenig:
Will be implemented differently after https://github.com/sifive/wit/pull/96

PR #84, opened 2019-06-13
add `wit foreach`

### LINKED ISSUE #83: Add a `wit foreach`
This should be analogous to `git submodule foreach`, it allows you to iterate on the wit packages and apply some operation to them. For example, one might want to create a pristine environment which involves deleting everything not under version control.

This relates to a more generic notion of `wit clean` to delete things that are not the packages themselves (with the possible exception of things installed by plugins like the scala plugin)

h/t @richardxia for the suggestion

aaronjanse:
fix #83

### DIFF
--- a/lib/main.py
+++ b/lib/main.py
@@ -71,8 +71,10 @@ def main() -> None:
 
     subparsers.add_parser('fetch-scala', help='Fetch dependencies for Scala projects')
 
-    foreach_parser = subparsers.add_parser('foreach', help='run a command in each package\'s working directory')
+    foreach_parser = subparsers.add_parser('foreach', help=("run a command in each "
+                                                            "package's working directory"))
     foreach_parser.add_argument('shell_command')
+    foreach_parser.add_argument('--dry-run', dest="dry_run", action="store_true")
 
     args = parser.parse_args()
     if args.verbose > 3:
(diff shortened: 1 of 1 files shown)


lib/main.py
```diff
@@ -302,6 +308,14 @@ def fetch_scala(ws, args, agg=True) -> None:
         scalaplugin.fetch_ivy_dependencies(files, install_dir, ivy_cache_dir)
 
 
+# NOTE: we could run commands in parallel but the potential for unexpected behaviour
+# may not be worth it
+def foreach(ws, args) -> None:
+    pkg_folders = [f.parent for f in ws.path.glob('*/wit-manifest.json')]
```
samh-sifive: I think this is a useful feature, but I wonder about a couple things here --

1. As written this will only operate on packages that contain a `wit-manifest.json` file, but not all packages necessarily have this file (they may not have any dependencies)

2. This may operate on packages that are no longer part of the workspace dependency hierarchy. Maybe that's OK? Maybe not? Not sure...

I feel like it may be safer to get a canonical package list by running the resolution algorithm (or maybe just reading `wit-lock.json`. I think we also should have a `-n` option here that just says which packages *would* be processed, without actually running the command in them. Like:

```
Skipping execution because of -n. Would have run:
cd mypackage1 && git tag foo
cd mypackage2 && git tag foo
cd mypackage3 && git tag foo
```

aaronjanse: You're right. I think this is pending on #33 (package resolution without changing workspace files), which is pending on a finalized `update` function.

richardxia:
Replaced by https://github.com/sifive/wit/pull/213.

PR #85, opened 2019-06-13
add `wit clean`

aaronjanse:
I don't know if there is much of a demand for a `wit clean` command, but here's some code.

It currently only removes workspace subdirectories that aren't packages.
Should it also `git reset --hard HEAD` packages? If it does, people could lose important work to their uncommitted manifest changes.

Or maybe we do `git reset --hard HEAD` but confirm first if that includes resetting manifests?

### DIFF
--- a/lib/main.py
+++ b/lib/main.py
@@ -24,6 +24,8 @@ from typing import cast, List  # noqa: F401
 from lib.common import WitUserError, error
 from lib.gitrepo import GitRepo
 import re
+from glob import glob
+import pathlib
 
 log = getLogger()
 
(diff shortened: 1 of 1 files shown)


lib/main.py
```diff
@@ -319,3 +326,14 @@ def version() -> None:
         version = re.sub(r"^v", "", version)
 
     print("wit {}".format(version))
+
+
+def clean(ws, args) -> None:
```
samh-sifive: See my comment in the `wit foreach` PR -- Unfortunately `wit-manifest.json` isn't a reliable indicator of package-ness. Iterating through the packages from the wit resolution algorithm is the safest way to do this.

PR #87, opened 2019-06-14
add test and fix for unnecessary ancestry error

aaronjanse:
The following no longer fails: (is this a good thing?)

(the numbers are years)
```
top:2019
├─ bar:2018
│  └─ xyz:2010
└─ foo:2017
   └─ xyz:2011
```


t/bad_ancestry_error.t
```diff
@@ -0,0 +1,57 @@
...
+cat << EOF | jq . > foo/wit-manifest.json
+[
+    { "commit": "$xyz_commit_2", "name": "xyz", "source": "$PWD/xyz" }
+]
+EOF
+git -C foo add -A
+git -C foo commit -m "add xyz:2"
+foo_commit=$(git -C foo rev-parse HEAD)
+
+# Set up repo foo
```
samh-sifive:
```suggestion
# Set up repo bar
```

lib/workspace.py
```diff
@@ -143,7 +143,8 @@ def update(self):
 
                 # 4. If the repo has a selected version, fail if that version
                 # does not include this tuple's hash
-                if not repo.is_ancestor(commit, selected_commit):
+                if not (repo.is_ancestor(commit, selected_commit)
```
samh-sifive: I had to think through the algorithm again to wrap my head around why the code was originally written as it was. So this is an interesting question that unfortunately rests on the /intention/ of the repo authors.

If the author of `bar:2018` explicitly specified `xyz:2010` despite newer versions existing (which is the only way that `foo:2017` could have a newer version: `xyz:2011`), is it safe to override that requirement with the newer version?

The commit-time based ordering of the repos is intentional and this somewhat defeats that. I'm trying to think through a case where this causes problems; if `xyz` has a set of dependencies that need to be processed, we could end up in a weird state... Curious to hear @jackkoenig and @terpstra thoughts on this one.

t/bad_ancestry_error.t
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env bash
+
+# (the numbers are years)
+#
+# top:2019
+# ├─ bar:2018
+# │  └─ xyz:2010
+# └─ foo:2017
+#    └─ xyz:2011
```
jackkoenig: It would be helpful if the numbers after the `:` here corresponded to the numbers in the commit messages below, just for understanding the test

lib/workspace.py
```diff
@@ -143,7 +143,8 @@ def update(self):
...
+                        or repo.is_ancestor(selected_commit, commit)):
```
aaronjanse: We need to run some special code in this scenario.

> If the author of bar:2018 explicitly specified xyz:2010 despite newer versions existing (which is the only way that foo:2017 could have a newer version: xyz:2011), is it safe to override that requirement with the newer version?

I think a warning would be a good thing to put here.

We also need to update the package to the newer version requested. When we do this, we might leave behind a dependency no longer needed (not the end of the world, but still undesired behavior).

lib/workspace.py
jackkoenig: Alright I discussed with @terpstra and the algorithm *should* support this test case, the implementation of the algorithm is slightly wrong. I haven't read the code, but in the example:
```
top:2019
├─ bar:2018
│  └─ xyz:2010
└─ foo:2017
   └─ xyz:2011
```
We visit `top` (pushing `bar` and `foo` into the priority queue), then visit `bar` and push `xyz:2010` to the priority queue, then we visit `foo` and push `xyz:2011` to the priority queue, then `xyz:2011` is newer than `xyz:2010` so we visit `xyz:2011`.

This fix is wrong, @aaronjanse is investigating

aaronjanse: The issue is that in order to know the timestamp of a commit, we have to clone the repo.

We're cloning `xyz` from `bar` to figure out whether we should explore it before or after `foo`.

```bash
# FIXME: This should clone to a temporary area. If this were
# fixed we could get rid of the source_map dictionary hack
```
If we clone to a temporary directory, as suggested in the existing code comment, we would have seen that `xyz:2010` is older _before_ cloning it to the workspace.

richardxia: Sorry for the drive-by commenting, since I have no context on this codebase, but you can't make use of git commit timestamps for ordering since interactive rebasing allows one to create a git commit graph that does not match the chronological ordering. Not to mention that the `--date` option to `git commit` allows one to override the date of a commit.

This is something I've noticed that even GitHub gets wrong, since I'll sometimes reorder my commits before submitting a pull request, and GitHub will display them in chronological order, not commit history order. This is most noticeable when the tip of the branch has a commit date that is older than another commit in the history, and the CI status icon appears on a commit that is not at the bottom of the page.

PR #88, opened 2019-06-14
fix confusing error messages when updating package

aaronjanse:


jackkoenig:
Alternative suggestion:
```
[ERROR] Cannot update package 'foo':
It has been pulled in as a dependency, but has not actually been added to the workspace.
```
Also note the `:` on the first line which @samh-sifive requested

aaronjanse:
> It has been pulled in as a dependency, but has not actually been added to the workspace.

The issue with "has not... been added" is that it doesn't make sense in the scenario when a package was added but later manually deleted by the user.

How about:
```
[ERROR] Cannot update package 'foo':
It has been added as a dependency but is not present in the workspace.
```

We might even want to add a note:
```
`wit update` might be able to resolve this.
```

jackkoenig:
> `wit update` might be able to resolve this.

This is not quite the right message. When you get this error, you either need to be updating the package as a dependency in another package, or you can `add-pkg` instead.

jackkoenig:
Yeah I'm still not sure about the best wording for the error

PR #89, opened 2019-06-14
Make ancestry error message more helpful

aaronjanse:
New error message:
```
Ancestry issue:
'foo' and 'bar' both depend on 'xyz':
    foo::b772b5f6 depends on xyz::9b0c2643
    bar::3106646d depends on xyz::f861f413

Although xyz::f861f413 is newer than xyz::9b0c2643,
xyz::f861f413 is not a descendent of xyz::9b0c2643.

Therefore, there is no guarantee that the dependee needed by bar::b772b5f6 will be satisfied by the dependee needed by bar::3106646d.
```


lib/workspace.py
```diff
@@ -14,8 +14,63 @@
...
+                "Although {child_name}::{newer_child_hash} is newer than "
+                "{child_name}::{older_child_hash},\n{child_name}::{newer_child_hash} is not "
+                "a descendent of {child_name}::{older_child_hash}.\n\n"
+                "Therefore, there is no guarantee that "
+                "the dependee needed by {newer_parent_name}::{older_parent_hash} will be satisfied "
+                "by the dependee needed by {newer_parent_name}::{newer_parent_hash}."
+                "".format(
+                    parent_a_name=self.parent_a.name,
+                    parent_b_name=self.parent_b.name,
+                    parent_a_hash=self.parent_a.revision[:8],
```
samh-sifive: Make a `GitRepo.short_revision` method that returns `revision[:8]`

aaronjanse:
I cannot figure out how to hide the stacktrace for the NotAncestryError. Unless someone else knows how, I think we should merge this and figure out aesthetics later.

I've tried inheriting WitUserError and catching NotAncesterError specifically, etc. Maybe we ditch the error class and just send a string to lib.common.error?

jackkoenig:
Agreed on merging as is, we'll figure it out later

PR #90, opened 2019-06-15
fix implementation of resolution algorithm

aaronjanse:
`pop` returns the last item by default (!)

### DIFF
--- a/lib/workspace.py
+++ b/lib/workspace.py
@@ -137,7 +137,7 @@ class WorkSpace:
         while queue:
             # 3. Pop the tuple with the newest committer date. This removes from
             # the end of the queue, which is the latest commit date.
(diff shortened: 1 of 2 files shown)


aaronjanse:
Hmmm, I added a `log.info(commit_time)` statement, and the `update` function still doesn't seem to be going from newest to oldest

aaronjanse:
The function seems to work when we use `pop()` instead of `pop(0)`

I still don't know why were getting the ancestry errors we didn't expect.

aaronjanse:
Turns out that the `commit_time` variable is in seconds. Because of that, all the commits in our `bad_ancestry_error` test had the same timestamp in the `update()` function, even though the commits had a specific error.

PR #91, opened 2019-06-17
add wit inspect

### LINKED ISSUE #76: pretty-print dependency tree
~~Pending on #72.~~

I've written a pretty printer for dicts (only 21 LoC).

Once our `update` function is finalized, we can rewrite it to print a dependency tree once it's done. Later, we can add an option to run `update` in a `/tmp` directory, allowing users to check the dependency tree without touching their working directory.

aaronjanse:
Fix #76 

We could add colors later on to make the tree more readable

```
bar@9a045e56
└─ xyz@2a0b2c81->13dfbe03
foo@814e0e1f
├─ baz@8dc2af98
└─ xyz@13dfbe03
   └─ fizz@38ef24c0
```


aaronjanse:
I added a commit for generating `dot` files. If desired, I can make it a separate pull request, but I thought it'd be easier to put it here than deal with merging/rebasing later

lib/main.py
```diff
@@ -134,6 +139,13 @@ def main() -> None:
 
             elif args.command == 'fetch-scala':
                 fetch_scala(ws, args, agg=False)
+
+            elif args.command == 'inspect':
+                if args.dot or args.tree:
+                    inspect_tree(ws, args)
+                else:
+                    log.error("wit inspect must be run with the --tree flag")
+                    sys.exit(1)
```
jackkoenig: better to use the `error` method here in case we want to refactor error reporting: https://github.com/sifive/wit/blob/3c23d2217c5c23eb99a4e9b1e160fc33d9eb84b8/lib/common.py#L9

Also you should at least update this error message to mention that it can be run with `--dot` instead. Alternatively, you could look into using Argparse for this and make it know that one of these must be passed: https://stackoverflow.com/a/26992149/2483329, there are mutually exclusive groups and maybe you can make it mutually exclusive and required. Not sure if this is the best way, but worth considering

Issue #75, opened 2019-06-12
add field to manifest to specify repo name


aaronjanse:

aaronjanse:
Relevant: #7

aaronjanse:
Hmmm, should this override the name a package was requested as?
If so, it might surprise people when a package is suddenly renamed.

PR #96, opened 2019-06-20
Refactor codebase

### LINKED ISSUE #33: New command: wit resolve

### LINKED ISSUE #78: Add field to manifest to specify "replaces"
This is for renaming a repo/package. Related to #75 

From the perspective of the dependency resolution algorithm, it basically means that the replacing repo and the replacee repo(s) are the same repo and should be treated as such.

aaronjanse:
~~Note this it implements the resolution algorithm correctly. That is known to cause issues (#90)~~

I also included a demo of some additional detail in `wit status`. I can remove this from the branch if necessary.

Last thing: the resolve() function depends on being able to clone in workspace_root/.wit when we tell it not to touch the root workspace.

---

I split up `GitRepo` into 3 separate classes:
1. `GitRepo`: simply manage a package on disk (very little code changed)
2. `Package`: what's found in `wit-lock.json`
3. `Dependency`: what's found in `wit-workspace.json`

I think the best way to visualize this is by looking at a verbose graph of [block-pio-sifive](https://github.com/sifive/block-pio-sifive): (click to see it up close)

drawing

In the above visualization, the boxes are `Packages`, and the circles are the `Dependencies`.

The `wit-workspace.json` along with each `wit-manifest.json` request a `Dependency`. As the package resolution algorithm is run, similar `Dependencies` are group together to point to the same `Package`. That `Package` is the version that "won" (i.e. the latest version).

After running the resolution algorithm, we get a list of `Packages`. We save this list in `wit-lock.json`. Then, we optionally tell each `Package` to make its `GitRepo` checked out to the correct version.

The separation between the 3 classes allows us to do a few things:
1. Resolve without touching the workspace root (fix #33)
2. Rename `Packages` without confusing already-cloned `Dependencies` expecting the old name (fix #78 #75)
3. Compare resolved `Packages` to the `GitRepo` on disk, telling us exactly what running `wit update` would do (for `wit status`)
4. Scan the dependency tree for error reporting:
a. I already have some WIP code (in a different branch) for the Conflicting Source error that lists what packages depend on which source for a given Dependency name
b. Upon an ancestry error, we can tell the user what _chain_ of dependency links needs to be updated

---

Since this is a major rewrite and includes the changes made to `master` since `refactor-packages` branched off, I don't recommend using the GitHub GUI to merge. Rather, once we reach agreement, I recommend:

```
git checkout refactor-packages
git merge -s ours master
```


lib/wit/package.py
```diff
@@ -1,73 +1,149 @@
...
-
-if __name__ == '__main__':
-    import doctest
-    doctest.testmod()
+
+        non_hash = len(self.unresolved_revision) < 40
+
+        if non_hash and not download:
+            log.error("Cannot create a reproducible workspace!")
+            raise WitBug("Cannot resolve '{}' without permission to download".format(non_hash))
```
jackkoenig: Shouldn't we check if we've already checked out the repo before erroring here? What if someone asks for a non_hash for something already in the workspace?

lib/wit/package.py
```diff
@@ -1,73 +1,149 @@
...
-            msg = "'{}' is not a package in workspace at '{}'".format(name, ws.path)
-            raise NotAPackageError(msg)
-        return package
-
-
-if __name__ == '__main__':
-    import doctest
-    doctest.testmod()
+
+        non_hash = len(self.unresolved_revision) < 40
```
jackkoenig: What value is there in handling hashes differently than non_hashes? I'm a little uncomfortable having possibly incorrect logic since I have actually seen branch names longer than 40 characters

lib/wit/package.py
```diff
@@ -1,73 +1,149 @@
...
+    def status(self, lock):
+        if lock.contains_package(self.name):
+            if not self.in_root:
+                return "\033[93m(missing)\033[m"
+            elif self.revision != self.repo.get_latest_commit():
+                return "\033[35m(will be checked out to {})\033[m".format(
+                    self.short_revision())
+        else:
+            if not self.in_root:
+                return "\033[92m(will be added to workspace and lockfile)\033[m"
```
jackkoenig: I don't think it's necessary in this PR, but I think we should factor the colors out into methods. Furthermore, you don't want to use colors when not connected to a terminal (eg. when being piped or redirected). Turning colors off is also often exposed as a command-line option (like `--no-tty` in wake). You can check if connected to a terminal with `sys.stdout.isatty()`

lib/wit/package.py
aaronjanse: Upstream may have different tags than local.

1. Bob downloads package `foo@origin/master`
2. Foo's remote moves `origin/master` to a different commit
3. Bob runs `wit update-pkg foo::origin/master`
4. Bob gets the manifest of the old `origin/master`
5. Alice, who has not downloaded `foo` before, runs `wit add-pkg foo:origin/master`
6. Bob and Alice have different workspaces
7. The workspace is non-reproducible (!)

If we don't fetch when our unresolved_revision is a tag, and the remote moves a tag, a user with the repo already cloned would get a different workspace than a user who has not yet downloaded the package.

lib/wit/package.py
```diff
@@ -1,73 +1,149 @@
...
+        self.repo = None
+
+        self.find_source(repo_paths)
+
+        self.dependents = []
+
+    def short_revision(self):
+        rev = self.revision or self.unresolved_revision
+        if len(rev) >= 40:
+            rev = rev[:8]
```
jackkoenig: 8 is too short, it looks like 12-characters is a better number for short revisions https://blog.cuviper.com/2013/11/10/how-short-can-git-abbreviate/

lib/wit/package.py
aaronjanse: We may have downloaded an old version, but we need to have the latest version downloaded

lib/wit/package.py
jackkoenig: We can also just let git tell us because it guarantees uniqueness (although that doesn't guarantee anything about future collisions)

lib/wit/package.py
aaronjanse: Agreed. We should definitely replace that line with a git command to check whether a string is a commit hash or a ref

aaronjanse: Fixed by ee5ea630c24edd8a2d35346dcfc7c7a5d1c76ce7

lib/wit/package.py
aaronjanse: Good catch. Fixed by https://github.com/sifive/wit/pull/96/commits/25c8bc59995342748ddce096d12abd6ce7925352

lib/wit/package.py
aaronjanse: Agreed, this would be a good thing to have, and it probably belongs in the logger class somehow.

However, I don't think anyone is parsing wit output, so maybe either:
- we exclude colors from this PR
- OR we fix colors in a different PR

lib/wit/dependency.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def __repr__(self):
+        return "Dep({})".format(self.tag())
+
+    def short_revision(self):
+        if self.package and self.package.repo:
+            if self.package.repo.is_hash(self.specified_revision):
+                return self.package.repo.get_shortened_rev(self.specified_revision)
+            return self.specified_revision
+        return self.specified_revision[:8]
```
aaronjanse: The reason I added the `[:8]` here is because, when doing `wit inspect --tree`, the dependency tag is requested before the repo is loaded. Therefore, without the repo loaded, the Dependency cannot get the shortest unique string representing its hash.

This only applies to hashes before arrows in the tree output. Lone hashes (e.g. `foo:2891731`) and replacement hashes (`->19371239`) are still based on `git rev-parse --short`.

Should we just cut it off at 8 or should we try loading repos earlier in the process of `wit inspect --tree`?

aaronjanse:
https://github.com/sifive/wit/pull/96/commits/50dc4a76ad0160f2c6dbda2abb6b2d48ead07b4d

I added a "Damaged Packages" field to `wit status` to avoid an error that previously occurred (folder not found).

aaronjanse:
tl;dr there's a bug that needs a small commit to fix, but it's a little complex to explain.

I'll need more time to work on it tomorrow.

jackkoenig:
Sounds good, thanks for the update!

aaronjanse:
Sadly, the fix commit (https://github.com/sifive/wit/pull/96/commits/3f86a2ff3b68f090387c34723f226c609b15746d) is much larger than I would have liked, but I think it is necessary.

Previously, I was using Package as a dummy object between a Dependency and a GitRepo. Packages were not meant for this, but I was using them like this anyway.

Packages are now much more explicit about whether they are "resolved" as the winner, allowing us to link same-named Dependencies together without implying that one of those Dependencies is the "winner."

This is explained in the [Package docstring](https://github.com/sifive/wit/blob/3f86a2ff3b68f090387c34723f226c609b15746d/lib/wit/package.py#L18-L30):

> Packages are the "winner" of several dependencies of the same name. 
> These "winner" Packages are determined by WorkSpace.resolve and stored in the wit-lock.json
> 
> Before we know which dependency "won," we need a Package object to 
> link these same-named Dependencies together.
>
> This transitive state is indicated by Package.revision == None
>
> We need repo_paths here because we will use it to transform sources we need both for
> downloading repos while in the transitive state and for downloading the "winning" git repo

---

The fixed bug was not previously noticed because it only affected broken workspace afaik

lib/wit/package.py
jackkoenig: It's not just about parsing output, it's also about piping to a log file. Anyway, we can fix it in a follow on PR

lib/wit/dependency.py
```diff
@@ -0,0 +1,158 @@
...
+        for subdep in subdeps:
+            tree[subdep.get_id()] = subdep.crawl_dep_tree(wsroot, repo_paths, packages)
+        return tree
+
+
+def parse_dependency_tag(s):
+    # TODO Could speed up validation
+    #   - use git ls-remote to validate remote exists
+    #   - use git ls-remote to validate revision for tags and branches
+    #   - if github repo, check if page exists (or if you get 404)
```
jackkoenig: These comments are obsolete (at least with regards to this code). They might apply elsewhere, otherwise they should be removed

lib/wit/gitrepo.py
```diff
@@ -214,11 +227,12 @@ def _git_command(self, *args):
...
-            log.error("Command [{}] exited with non-zero exit status [{}]"
-                      .format(' '.join(proc.args), proc.returncode))
-            log.error("stdout: [{}]".format(proc.stdout.rstrip()))
-            log.error("stderr: [{}]".format(proc.stderr.rstrip()))
-            raise GitError(proc.stderr.rstrip())
+            msg = "Command [{}] exited with non-zero exit status [{}]\n".format(
+                  ' '.join(proc.args), proc.returncode)
+            msg += "stdout: [{}]\n".format(proc.stdout.rstrip())
+            msg += "stderr: [{}]\n".format(proc.stderr.rstrip())
+            msg += proc.stderr.rstrip()
```
jackkoenig: This adds stderr twice

aaronjanse: Fixed by https://github.com/sifive/wit/pull/96/commits/5acee8507ab3db97dc9ee6bab1a44b940c73ee0d

lib/wit/dependency.py
aaronjanse: I think that we should still eventually use `ls-remote` for resolving non-hash references. Should I move the comment to the top of `gitrepo.py`?

jackkoenig: Yeah that would make more sense

aaronjanse: Fixed by https://github.com/sifive/wit/pull/96/commits/ad53f1d1676df53397c121853a28620118f67f2c

lib/wit/common.py
```diff
@@ -16,3 +17,11 @@ class WitUserError(Exception):
     Supertype of user-input errors that should be reported without stack traces
     """
     pass
+
+
+# https://stackoverflow.com/a/845194
+def passbyval(func):
+    def new(self, *args):
```
jackkoenig: You might consider implementing `**kwargs` as well since if anyone tries to call a function decorated with this using keyword arguments, it will just silently drop them

jackkoenig:
`wit inspect` crashes if you delete a repo, while `wit status` does correctly mark it missing.

One minor quibble about the `wit status` output:
```
Clean packages:
    freedom-devicetree-tools
    sifive-blocks
    environment-sanmateo-sifive
    api-firrtl-sifive
    api-generator-sifive
    berkeley-hardfloat
    wit
    api-chisel3-sifive
    chisel3
    api-scala-sifive
    block-pio-sifive
    soc-testsocket-sifive
    firrtl
    freedom-metal
    rocket-chip
Dirty packages:
Damaged packages:
    rocket-chip-wake
rocket-chip-wake (missing)
```

Could it just mention `rocket-chip-wake` once (indented under "Damaged packages") and put the `(missing)` there?

lib/wit/common.py
aaronjanse: fixed by https://github.com/sifive/wit/pull/96/commits/61d0d7b7dade20530295f3ecd26e9ad38446141c

lib/wit/main.py
```diff
@@ -325,9 +325,9 @@ def status(ws, args) -> None:
         for path in untracked:
             relpath = path.relative_to(ws.root)
             log.info("    {}".format(relpath))
-    if len(damaged) > 0:
-        log.info("Damaged packages:")
-        for package in damaged:
+    if len(missing) > 0:
+        log.info("Missing packages:")
+        for package in missing:
```
jackkoenig: Is renaming it right? Doesn't the category "damaged" include other possible problems?

aaronjanse:
> wit inspect crashes if you delete a repo, while wit status does correctly mark it missing.

Turns out `wit inspect` was always crashing, not just when missing a repo.
Fixed by https://github.com/sifive/wit/pull/96/commits/e281c6cb7609da4c840ce317ed82deb60fc52bd1

lib/wit/main.py
aaronjanse: No, currently they are the same thing.
The cleanest thing to do would be to call Package.status to get each package's (category, labels) (e.g. `("Damaged", ["missing"])`) but I'm not sure if you want that in this PR

jackkoenig: Okay I see. Yeah it's fine to do that later

lib/wit/dependency.py
```diff
@@ -0,0 +1,154 @@
...
+            'name': self.name,
+            'source': self.source,
+            'commit': self.specified_revision,
+        }
+
+    # used before saving to manifests/lockfiles
+    def resolved(self):
+        if self.package is None or self.package.repo is None:
+            raise Exception("Cannot resolve dependency that us unbound to disk")
+        return Dependency(self.name, self.source, self.resolved_rev())
```
jackkoenig:
```suggestion
    def resolved(self):
        rev = self.resolved_rev()
        return Dependency(self.name, self.source, rev)
```

jackkoenig: The diff shows up wrong, but the idea being that the error check is redundant with that in `resolved_rev`.

aaronjanse: Good catch. That diff did look weird so I fixed it with https://github.com/sifive/wit/pull/96/commits/f27443d7754a8c326cd8afb4f01ebbace2b4961a

PR #98, opened 2019-06-21
Add "bad ancestry error" test with commits with unique timestamps

aaronjanse:
EDIT: this disturbingly seems to fail CI

### DIFF
--- a/t/bad_ancestry_error.t
+++ b/t/bad_ancestry_error.t
@@ -28,6 +28,8 @@ git commit -m "xyz:2"
 xyz_commit_2=$(git rev-parse HEAD)
 cd ..
 
+sleep 1
+
 make_repo 'foo'
 cat << EOF | jq . > foo/wit-manifest.json
 [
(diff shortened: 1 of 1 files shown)


aaronjanse:
We could optionally error upon duplicate timestamps with:
```patch
diff --git a/lib/workspace.py b/lib/workspace.py
index 02ffba1..15d27ef 100755
--- a/lib/workspace.py
+++ b/lib/workspace.py
@@ -197,6 +197,10 @@ class WorkSpace:

             # Keep the queue ordered
             queue.sort(key=lambda tup: tup[0])
+            if queue and queue[-1][0] == commit_time:
+                log.error("Multiple commits with the same timestamp:")
+                log.error("{}::{} and {}::{}".format(reponame, commit[:8], queue[-1][2], queue[-1][1][:8]))
+                sys.exit(1)

             # 12. Go to step 3 (finish loop)
```

aaronjanse:
This is fixed by a branch I'm working on that does this for all tests, without the need for actually waiting for several seconds

PR #99, opened 2019-06-22
Support fetching the Scala compiler plugin

jackkoenig:
This involved moving some code over from [api-scala-sifive](https://github.com/sifive/api-scala-sifive)

This fixes a bug where the wit scala-plugin wasn't actually fetching all necessary dependencies, namely the Scala compiler bridge. Unfortunately, after conversations with Bloop's main developer, there is no good way to fetch this, so I wrote some sketchy Scala to do it: This includes 0.1.0 of https://github.com/sifive/scala-compiler-bridge-fetcher as a jar (it's < 8KB). Future work can improve on scala-compiler-bridge-fetcher, or actually improve Bloop to fix the issue where this is necessary.

This isn't great (well, mainly the logic in the scala-compiler-bridge-fetcher repo), but it works. It increases typical fetching size from ~90MB to ~130MB. This could be improved, which I tried to do, but then it didn't work for other versions of Scala so I just decided to :shipit:


jackkoenig:
I believe this fixes many issues people have been seeing, unable to write to their home directory, parallelism issues, etc. I think once this is in we should get ready to tag 0.10.0

PR #103, opened 2019-07-03
Error upon duplicate timestamps

aaronjanse:
(there will be a merge conflict, but it shouldn't be difficult to merge once we settle on #96)


aaronjanse:
I'll re-submit this PR when ready

PR #106, opened 2019-07-09
attach HEAD to named reference if possible

aaronjanse:
`wit update` et al should no longer detach `HEAD` if a named reference exists at the appropriate revision.


aaronjanse:
https://github.com/sifive/wit/pull/106/commits/7c412f7b9fcb7b71834fa5112d0237c06712ebb5
Sample output:
```
Checking out 'xyz' at 'bfe2eb1126c183fda96a872d190539e9470a4ed5'
Checking out 'bar' at 'testing' (1c508f798bcbafff9bb9e98d85571094020b6651)
Checking out 'foo' at 'b7754c325a3a8deaa97762b23029bcd65fa95726' (master, other_branch, tags/yay)
```

aaronjanse:
The build is failing because of:

https://github.com/sifive/wit/blob/7c412f7b9fcb7b71834fa5112d0237c06712ebb5/lib/wit/gitrepo.py#L216-L217

When we clone, we don't check out. Because of the above check, we never check out.

Fixed by https://github.com/sifive/wit/pull/106/commits/e4de90bc8c5f63f8547cbc0cc1bd149a04ab5762

lib/wit/gitrepo.py
```diff
@@ -183,8 +187,31 @@ def update_dependency(self, dep):
...
+                suggestions = ' ({})'.format(', '.join(rev_names))
+
+            if len(rev_names) != 1:
+                rev = self.revision
+                log.info("Checking out '{}' at '{}'{}".format(self.name, rev, suggestions))
+            else:
+                rev = rev_names[0]
+                log.info("Checking out '{}' at '{}' ({})".format(self.name, rev, self.revision))
+
+        proc = self._git_command("checkout", rev)
```
jackkoenig: I think the act of checking out should also be guarded by `if current_commit != self.revision` right? Otherwise this should still detach HEAD?

aaronjanse: > the act of checking out should also be guarded by `if current_commit != self.revision` right?

After `GitRepo.clone`, `self.revision == current_commit` even though we aren't checked out. If we added the above check, `GitRepo.clone ; GitRepo.checkout` would never checkout.

Afaik, Git doesn't have a special "not checked out" state; it treats an un-checked-out repo exactly the same as it would if you ran `git checkout ; rm -rf ./*`

> Otherwise this should still detach HEAD?

Only when there are multiple references pointing to the same commit and `HEAD` points to one of those references. I'll fix this (good catch!)

aaronjanse: Fixed by https://github.com/sifive/wit/pull/106/commits/36acd4a0dc4fb3079f41e014e80a1059bfb7ed03

jackkoenig:
This looks great, but can you add a test? I'd like to see it test the following:

1. If a branch is specified on `wit init -a` or `wit add-pkg`, that branch will be checked out

2. wit doesn't do a checkout if HEAD matches the commit

This could be tested by making sure there are multiple branches pointing to the same commit, but so long as one is currently checked out it should stay checked out across a `wit update`

3. It doesn't check anything out if it's ambiguous but makes suggestions

Like in (2), multiple branches for the same commit, but make sure to check out some other commit before running `wit update`

aaronjanse:
> 1. If a branch is specified on `wit init -a` or `wit add-pkg`, that branch will be checked out

This might be difficult/impossible. Wit needs to resolve local branches into hashes in order to generate a manifest that can be shared with other people.

The branch name would be given to `update-pkg` but by the time `wit update` (which checks stuff out) is run, there will be no record of what branch was originally requested

---

> 2. wit doesn't do a checkout if HEAD matches the commit
> 3. It doesn't check anything out if it's ambiguous but makes suggestions

Fixed by https://github.com/sifive/wit/pull/106/commits/74635a113c735457611e80acbfab078d055a5df7

> This could be tested by making sure there are multiple branches pointing to the same commit, but so long as one is currently checked out it should stay checked out across a `wit update`

Fixed by https://github.com/sifive/wit/pull/106/commits/29858ae5429522935a0f599a07c8e54ae8dddd75

aaronjanse:
https://github.com/sifive/wit/blob/74635a113c735457611e80acbfab078d055a5df7/t/wit_head_attachment.t#L49-L53

I used `md5` because 30 minutes of DuckDuckGo (and even Google Search) got me absolutely nowhere regarding how to compare strings with spaces (which worked in my shell but not in the test file for some reason)

jackkoenig:
> This might be difficult/impossible. Wit needs to resolve local branches into hashes in order to generate a manifest that can be shared with other people.

Fair enough, perhaps worth revisiting someday but certainly not within the current implementation

PR #116, opened 2019-07-11
allow update-dep without internet access

### LINKED ISSUE #111: update-dep should not fail without internet access
`wit update-dep` will fail when it tries to `git fetch` if there is no internet access. Ideally `update-dep` should work without internet.

aaronjanse:
fixes #111 

I refactored the existing logic because the conditional statement was getting quite large.
Note the comment on the deleted code when considering how the code was rewritten.


aaronjanse:
This should be generalized to `load` only downloading for remote branches, but not for local branches that are already present

PR #119, opened 2019-07-16
quickly describe resolution algorithm

aaronjanse:
This is different from what I originally proposed, but I think it works better with what already exists in the readme?


README.md
```diff
@@ -14,6 +14,9 @@ Each package may optionally contain a wit-manifest.json file which defines other
 Wit resolves this hierarchy of dependencies and generates a flattened directory structure in which each package
 may exist only once.
 
+## How does Wit deduplicate packages?
+When multiple versions of the same package are requested, Wit chooses the latest version, making sure the latest version's commit is a descendant of the rest of the requested versions' commits.
```
jackkoenig:
```suggestion
When multiple versions of the same package are requested, Wit chooses the latest requested version, making sure the selected version's commit is a descendant of every other requested versions' commits.
```

PR #120, opened 2019-07-17
allow resolve to be interrupted

### LINKED ISSUE #35: Trap ctrl-c
ctrl-c shoudl print a warning that the workspace will be corrupt. Hit ctrl-c again to really kill.

### LINKED ISSUE #36: Support Resuming an Incomplete `wit init`
It's easy to make typos or other things that will cause a workspace to be partially created, then if you fix your command it errors because the workspace already exists.

One particular example I've run into is when you try to create a workspace that violates the ancestor requirement; it'll error and I have to nuke the workspace and then start all of the cloning over again

aaronjanse:
partially fixes #35 
fixes #36


aaronjanse:
Pending on #135

PR #124, opened 2019-07-17
auto-resolve conflicting sources

aaronjanse:


aaronjanse:
I'm not sure why this is failing CI. It passes on my machine.

aaronjanse:
I'm brainstorming how we'll roll out breaking changes. I think some users might be motivated to update to a version that auto-resolves source conflicts, so maybe we pair this with the release of something having to do with the upcoming `replaces` breaking change

PR #128, opened 2019-07-18
cleanup update and add

aaronjanse:
This is a sibling bugfix PR of #118.

NB: this PR includes a minor info stdout change


lib/wit/main.py
```diff
@@ -257,20 +264,18 @@ def update_dep(ws, args) -> None:
...
     # make sure the package is already in the cwd's manifest
     if not manifest.contains_dependency(req_dep.name):
         log.error("'{}' does not depend on '{}'".format(cwd_dirname, req_pkg.tag()))
         sys.exit(1)
 
+    manifest_dep.specified_revision = req_pkg.revision
+
     log.info("Updating to {}".format(req_dep.resolved_rev()))
-    manifest.replace_dependency(req_dep)
+    manifest.replace_dependency(manifest_dep)
```
aaronjanse: The idea is that, instead of replacing the item entirely, we update the existing item.

In the future, I'm thinking of doing a small refactor of this workflow, but I think verbosity shouldn't be a problem in the meantime

PR #132, opened 2019-07-19
make "commit not found" error message more helpful

aaronjanse:


lib/wit/gitrepo.py
```diff
@@ -15,7 +15,7 @@ class GitError(Exception):
     pass
 
 
-class GitCommitNotFound(WitUserError):
+class GitCommitNotFound(Exception):
```
jackkoenig: Why change this from `WitUserError` to `Exception`? The idea of `WitUserError` was that we could unify how we catch and give nice error messages for user errors. Generally I'd like to avoid having to sprinkle `try` `catch` all over the code base:
```python
        try:
            # something
        except GitCommitNotFound:
            error("Could not find commit or reference '{}' in '{}'"
                  "".format(dep.specified_revision, dep.name))
```
I would probably change it (if possible) to have the part changing from `GitError` to `GitCommitNotFound` include information about what went wrong in the Exception, and then just catch the exception in `main` instead (assuming you can get the info you need at that place).

aaronjanse: The idea is that, from inside GitRepo, we're not sure whether GitCommitNotFound was caused by incorrect user input or a bug in Wit.

We default to is being a bug in Wit (thus `Exception`) but catch it in certain places where GitCommitNotFound is known to be the user's fault (e.g. when resolving an argument to `add-dep`).

An example of a Wit bug that we could catch is the Wit implementation improperly mixed tags and hashes. If that was the case, we'd want a bug report with a stack trace instead of telling the user not to report because it's a user error.

jackkoenig: I agree with that, but I think it's still cleaner to have unified error reporting (in main.py rather than sprinkled throughout, better to have a single point of exiting the program), and rather than catching and turning into errors, we could just have a function that transforms `GitCommitNotFound` into `WitUserError` with the correct error message (the type of the exception is irrelevant since we don't report it)

aaronjanse: > I think it's still cleaner to have unified error reporting (in main.py rather than sprinkled throughout, better to have a single point of exiting the program)

I agree.

Is this fixed by https://github.com/sifive/wit/pull/132/commits/4edb1c170206a36a20c923ca3c44876bebb042b9 ?

I don't like the duplication of the error message, but at the same time I don't want to pass the dependency name to `GitRepo.get_commit` because that'd mean I'd have to pass the name to every function that calls `get_commit`.

PR #135, opened 2019-07-22
auto-resolve sources

### LINKED ISSUE #105: detect similar github sources
We don't want to throw a Conflicting Sources error for sources that are different strings but point to the same thing. We could write special code for github repos such that wit understands that the following are identical:

```
https://github.com/sifive/wit
https://github.com/sifive/wit.git
git@github.com/sifive/wit
git@github.com/sifive/wit.git
```

The only problem is if the user has a preference of whether they clone over https vs ssh.
Maybe we shouldn't expect users to have both a password token and a ssh key?

### LINKED ISSUE #121: Automatically resolve "conflicting sources"
Wit will always find a source with the newer commits first (due to the nature of its resolution algorithm). Why don't we just make sure the commit with the "conflicting source" is a descendant of the commit we already found in the package with the same name?

This would make `--repo-path` needed much less often, which might make Wit much more beginner-friendly.

### LINKED ISSUE #122: handle non-existent repository source
Current error output when running `repo_path.t`:
<details><pre>
Traceback (most recent call last):
  File "/nix/store/jk588wcj1wg51kv2af3rb2j7lkl8h9b0-python3-3.5.7/lib/python3.5/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/nix/store/jk588wcj1wg51kv2af3rb2j7lkl8h9b0-python3-3.5.7/lib/python3.5/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/Users/aaronj/wit/lib/wit/__main__.py", line 11, in <module>
    main()
  File "/Users/aaronj/wit/lib/wit/main.py", line 114, in main
    create(args)
  File "/Users/aaronj/wit/lib/wit/main.py", line 185, in create
    update(ws, args)
  File "/Users/aaronj/wit/lib/wit/main.py", line 347, in update
    packages, errors = ws.resolve(download=True)
  File "/Users/aaronj/wit/lib/wit/workspace.py", line 117, in resolve
    source_map, packages, queue, errors)
  File "/Users/aaronj/wit/lib/wit/common.py", line 11, in new
    return func(self, *copy_args, **copy_kwargs)
  File "/Users/aaronj/wit/lib/wit/dependency.py", line 30, in resolve_deps
    subdep.package.load_repo(wsroot, download, subdep.specified_revision)
  File "/Users/aaronj/wit/lib/wit/package.py", line 88, in load_repo
    self.repo.download()
  File "/Users/aaronj/wit/lib/wit/gitrepo.py", line 75, in download
    self.clone()
  File "/Users/aaronj/wit/lib/wit/gitrepo.py", line 86, in clone
    self._git_check(proc)
  File "/Users/aaronj/wit/lib/wit/gitrepo.py", line 239, in _git_check
    raise GitError(msg)
wit.gitrepo.GitError: Command [git clone --no-checkout /Users/aaronj/wit/t/foo /Users/aaronj/wit/t/myws/.wit/foo] exited with non-zero exit status [128]
stdout: []
stderr: [fatal: repository '/Users/aaronj/wit/t/foo' does not exist]
</pre></details>

---

Maybe we should tell Wit users to report all stacktraces they see. For each stacktrace, we could either improve the error logging and/or fix a bug.

aaronjanse:
I merged #133 into this PR because debugging is much easier when paired together.

Fixes #122
Fixes #121  
Fixes #105  

I merged `load_package` and `load_repo` into method `load` implemented in both Package and Dependency. This eliminated a lot of duplicate code.

For the refactor, I had to merge the following bugfix PRs:
#118  
#128  

Area of particular interest:
https://github.com/sifive/wit/blob/cfc334b7631b987e9532e82a81f98428e793819b/lib/wit/package.py#L132-L144

---

In general, I strive to practice "one commit/branch does one thing," but bundling a few things into this PR should have made it easier to review.


aaronjanse:
I just fixed linter problems. I'm not sure why the tests fail on CI (they pass locally).
This should be ready to review

aaronjanse:
Fixed typo. Side note, we should be testing `wit status` https://github.com/sifive/wit/issues/136

t/repo_path_conflict_resolvable.t
```diff
@@ -32,7 +32,7 @@ prereq "off"
 wit init myws -a $PWD/main_repo
 
 # Should fail because of conflicting paths for foo
```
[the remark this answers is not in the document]
aaronjanse: > Can you refactor this test to be a good example of what support you're adding

Will do.

> We also want to be really careful that the on-disk clone of a multi-source repo has the "winning" remote set to `origin`

Agreed. See:
https://github.com/sifive/wit/blob/cfc334b7631b987e9532e82a81f98428e793819b/lib/wit/package.py#L138-L144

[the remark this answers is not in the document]
aaronjanse: > Can you refactor this test to be a good example of what support you're adding?

Is this fixed by https://github.com/sifive/wit/pull/135/commits/f3fd436b81854ebdc3753180b9d9f5279fd54f86 ?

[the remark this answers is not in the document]
jackkoenig: Yep, nice test :)

lib/wit/dependency.py
```diff
@@ -40,16 +40,21 @@ def resolve_deps(self, wsroot, repo_paths, download, source_map, packages, queue
...
 
             if subdep.name in source_map:
                 if subdep.package.resolve_source(subdep.source) != source_map[subdep.name]:
-                    log.error("Dependency [{}] has multiple conflicting paths:\n"
-                              "  {}\n"
-                              "  {}\n".format(subdep.name, subdep.source,
-                                              source_map[subdep.name]))
-                    sys.exit(1)
+                    if not packages[subdep.name].repo.has_commit(subdep.specified_revision):
+                        error(("Two dependencies have the same name "
```
jackkoenig: In general, I'd rather we do all erroring from the same location in `main` where we catch `WitUserError`. Concretely, I think this means turning other `error` calls into throwing `WitUserError`, and just setting the message such that it prints the same message.

Why do I think this is useful? It ensures that we can catch and correct or rethrow errors where possible (rather than `sys.exit` occurring deep in the call stack), it unifies the handling of errors such that we can eventually assign return codes, and it enables unit testing since we can catch the exceptions.

jackkoenig: I would appreciate opinions on this from @richardxia since he's more experienced in Python than me

lib/wit/gitrepo.py
```diff
@@ -32,44 +41,46 @@ class GitRepo:
...
+        self.path.mkdir()
+        proc = self._git_command("clone", "--no-checkout", source, str(self.path))
+        try:
+            self._git_check(proc)
+        except GitError:
+            if proc.stderr.startswith("fatal: repository") and \
+               proc.stderr.endswith("does not exist\n"):
+                raise BadSource(self.name, source)
+            else:
+                raise
```
jackkoenig: Does this `raise` mean it will reraise the same error as before or does it replace it with a generic, information-less exception?

lib/wit/package.py
```diff
@@ -55,16 +58,17 @@ def add_dependent(self, dep):
...
         """
 
-        needed_commit = needed_commit or self.revision
+        source = self.resolve_source(source) or self.resolve_source(self.source)
+        revision = revision or self.revision
 
-        if needed_commit is None:
-            raise Exception("Cannot load repo for unknown commit.")
+        assert revision is not None, WitBug("Cannot load repo for unknown commit.")
+        assert source is not None, WitBug("Cannot load repo for unknown source.")
```
jackkoenig: Does this throw a `WitBug`? I wasn't aware of this syntax, nice!

lib/wit/package.py
```diff
@@ -122,14 +127,27 @@ def add_dependency(self):
...
+            if self.repo.path.parts[-2] == '.wit':
+                self.repo.set_origin(self.source)
+            else:
+                log.warn("Package '{}' wants a different git remote origin.\n"
+                         "Origin is currently:\n  {}\n"
+                         "Package '{}' wants origin:\n  {}\n"
+                         "Please manually update the origin with:\n"
+                         "  git -C {} \\\n    remote set-url origin {}"
+                         "".format(self.name, current_origin, self.name,
+                                   wanted_origin, self.repo.path, wanted_origin))
```
jackkoenig: So this warning will occur if you have a workspace, and then change it in such a way that we introduce a dependency on a newer version of a repo with a different remote?

jackkoenig: I think a test of this behavior would be good

lib/wit/gitrepo.py
richardxia: It reraises the same exception, and this is probably the more idiomatic form because in Python 2.x, this would be the easiest way of ensuring that the stack trace is preserved as well.

lib/wit/package.py
richardxia: I just tried this out locally with a toy example, and no, it does not. It still throws an `AssertionError`. The [Python docs](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement) say that this is equivalent to `raise AssertionError(WitBug("foo"))`.

lib/wit/dependency.py
richardxia: I agree with not overusing a helper function that calls `sys.exit()`, since that makes it impossible to do any handling of errors.

lib/wit/package.py
aaronjanse: Correct. However, WitBug being the type of the error message for AssertionError helps the user know that if they see that stacktrace, it's something they should report

lib/wit/package.py
aaronjanse: > So this warning will occur if you have a workspace, and then change it in such a way that we introduce a dependency on a newer version of a repo with a different remote?

Exactly.

> I think a test of this behavior would be good

 Will do.

lib/wit/package.py
aaronjanse: Better idea: I'll make the assertion error message a string, catch AssertionError in `main.py`, then wrap it with `WitBug` before reraising

aaronjanse:
I fixed the way that `source[::rev]` tags are parsed from user input. Area of interest:
https://github.com/sifive/wit/blob/495dd00350b51f5e25a2f4617b8c8f1716a8e142/lib/wit/main.py#L200-L211

Relevant tests:

[wit_add_pkg_already_cloned.t](https://github.com/sifive/wit/blob/5670099b45988e16c765ed696045231684de3b5d/t/wit_add_pkg_already_cloned.t) 
[wit_add_pkg_already_dep.t](https://github.com/sifive/wit/blob/495dd00350b51f5e25a2f4617b8c8f1716a8e142/t/wit_add_pkg_already_dep.t)

---

TODO:
- [ ] add test for BadSource error
- [ ] change `error` -> `raise WitUserError`
- [ ] improve assertion error code

lib/wit/package.py
aaronjanse: > I think a test of this behavior would be good

Fixed by https://github.com/sifive/wit/pull/135/commits/cab246f18a394d8cbcc47e0fdeec44793781d074

lib/wit/dependency.py
aaronjanse: Fixed by https://github.com/sifive/wit/pull/135/commits/be5e8ce9f1eb50e877dd8db153afa738a98a326d

t/sources_conflict_unresolvable.t
```diff
@@ -8,10 +8,15 @@ prereq "on"
 make_repo 'foo'
 foo_commit=$(git -C foo rev-parse HEAD)
 
+sleep 1  # prevent duplicant commit hash on Travis (crazy)
```
aaronjanse: Craziness

aaronjanse:
```suggestion
sleep 1  # prevent duplicate commit hash on Travis (crazy)
```

PR #138, opened 2019-07-23
handle manifests with replaces field

aaronjanse:
This does not (yet???) write manifests in the new format


aaronjanse:
Area of interest:
https://github.com/sifive/wit/blob/f2d3d23b3a7fbc7cfb6b5d3eb067554d58da9701/lib/wit/workspace.py#L177-L180

PR #144, opened 2019-07-30
implement interruptible resolve

### LINKED ISSUE #35: Trap ctrl-c

aaronjanse:
Fixes #35

### DIFF
--- a/t/wit_interrupted_update.t
+++ b/t/wit_interrupted_update.t
@@ -1,0 +1,29 @@
+#!/bin/sh
+
+. $(dirname $0)/regress_util.sh
+
+# Now create a workspace from bar
+wit init myws
+cd myws
+
+wit add-pkg https://github.com/sifive/block-pio-sifive
+
+set -x
+
+# brew install coreutils
+timeout 3 wit update
+
+ls block-pio-sifive
+
+check "packages should not be checked out when update is interrupted" [ $? -ne 0 ]
+
+wit update
+
+ls block-pio-sifive
+
+check "packages should be checked out after update is resumed" [ $? -eq 0 ]
+
+set +x
+
+report
+finish
--- a/t/wit_interrupted_update.t
+++ b/t/wit_interrupted_update.t
@@ -11,17 +11,16 @@ wit add-pkg https://github.com/sifive/block-pio-sifive
 set -x
 
 # brew install coreutils
-if [ "$CI" = "true" ]; then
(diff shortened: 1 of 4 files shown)


aaronjanse:
Travis is failing but local test work. Nice.

t/wit_interrupted_update.t
```diff
@@ -0,0 +1,33 @@
...
+wit init myws
+cd myws
+
+wit add-pkg https://github.com/sifive/block-pio-sifive
+
+set -x
+
+# brew install coreutils
+if [ "$CI" = "true" ]; then
+    git config --global url."https://github.com/".insteadOf 'git@github.com:'
```
aaronjanse: :tada:

aaronjanse:
In case it isn't already abundantly clear, this is gonna need a squash & merge as opposed to a normal merge.

jackkoenig:
@richardxia what do you think about this? I like the feature but am unsure about the implementation, I suspect that `sys.exit(1)` upon a `KeyboardInterrupt` is not generally the way to do this but it's fine, seems to work okay.

richardxia:
> @richardxia what do you think about this? I like the feature but am unsure about the implementation, I suspect that `sys.exit(1)` upon a `KeyboardInterrupt` is not generally the way to do this but it's fine, seems to work okay.

I am not familiar with this either, but I've found some stuff on the internet that suggests that when a POSIX process is killed by a signal, it should have an exit code of 128 + the signal number. Since SIGINT is 2, the exit code should be 130. This is in fact what a bash script will return with if you SIGINT to exit it early. See https://unix.stackexchange.com/a/99143

Also, does this PR actually implement what was described in #35? I don't think this actually does anything beside force the exit code to be 1 when you catch a `KeyboardInterrupt`. If you do want to warn the user and require a second SIGINT to actually terminate the process, you need to have a much more complicated exception handler where you don't immediately exit. See https://stackoverflow.com/a/18115530/7433423 for an example that does this.

I'm actually not really sure exactly what happens in Python at the threading level when you set a signal handler like that.

Anyway, I'm actually not not really convinced that this PR does anything, since I don't think the `try; except KeyboardException` block actually does anything in terms of the control flow other than not printing a stack trace and exit 1.

richardxia:
I think the codebase has probably diverged too much since when this was originally opened, so I'm going to close it out.

PR #152, opened 2019-08-22
Add Wit Dockerfile

LeytonHo:
Wit Docker image corresponding to `wit.Dockerfile` can be found on Docker Hub. For every new Wit release, the Docker image is rebuilt, tagged, and pushed to Docker Hub.


docker/build_image.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+docker login -u $USERNAME --password $PASSWORD
+export TAG=`if [ "$TRAVIS_BRANCH" = "master" ]; then echo "latest"; else echo $TRAVIS_BRANCH ; fi`
+docker build --file=docker/wit.Dockerfile --tag=sifive/wit:$TAG .
+if [ "$TRAVIS_TAG" != "" ]; then docker push sifive/wit; fi
```
richardxia: It's a little bit misleading to have the script named `build_image.sh`, since it doesn't imply that an image would also get pushed, and one might attempt to run that without realizing that it will also cause a change in Docker Hub.

I think it would be a good idea to have the push go in a separate script. The `docker login` would also go with that, since it's only needed for pushing. You can have Travis only run that script as part of the `deploy:` top level key, which only runs when all the tests actually pass. You can also set rules so that the `deploy:` step only runs on certain branches/tags so that we don't attempt to do a push to Docker Hub on every single pull request.

See https://docs.travis-ci.com/user/deployment/script/ for more information.

docker/wit.Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM ubuntu:xenial
+
+RUN apt-get update && apt-get install -y \
+   openjdk-8-jdk \
+   software-properties-common \
+   git  
+
+RUN git clone --branch v0.10.1 https://github.com/sifive/wit.git
```
richardxia: Since this _is_ the `wit` repository, you should just do a copy from the current directory.

jackkoenig:
How hard is it to delete something from DockerHub? Once this is merged can we do a testrun where I tag a commit, it pushed, then I delete the tag and docker image?

jackkoenig:
Maybe that's a bad idea and the "test case" will just be `0.11.0`

LeytonHo:
> How hard is it to delete something from DockerHub? Once this is merged can we do a testrun where I tag a commit, it pushed, then I delete the tag and docker image?

It's very easy to delete a tag from Docker Hub so a test run should be fine.

docker/wit.Dockerfile
jackkoenig: Furthermore, you should use `make install` because it removes unneeded stuff like the tests. You can do `make install PREFIX=/path/to/install` to set the directory where it puts the code. Note that it installs to a directory with the version like `/path/to/install/0.10.1`. You can override the version with `version=whatever` if needed.

docker/push_image.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+set -euxo pipefail
+
+docker login -u $DOCKER_USERNAME -p $DOCKER_PASSWORD
+docker push sifive/wit
```
richardxia: I think you need to set the tag/version here or else I believe it'll push `latest`.

.travis.yml
```diff
@@ -1,8 +1,16 @@
...
+language: python
 python:
     - '3.5'
 
+deploy:
+  provider: script
+  script: ./docker/push_image.sh
+  on:
+    tags: true
+
```
richardxia: Just as a personal style, I like putting the `deploy` key below the `jobs` key, since the `deploy` stuff only runs after all the other steps pass.

PR #174, opened 2019-11-08
Upgrade coursier to 2.0.0-RC4-1 and use binary by default

jackkoenig:
fetch-scala now has --jar option to force download of jar
Binary distributions are available for 64-bit Linux and 64-bit MacOS

This PR also introduces sha256 checking for the coursier download as an added safety mechanism.

We use coursier in `wit` for Scala dependency fetching as well as in `api-scala-sifive`'s wake rules for dependency resolution. We have to invoke coursier for every unique set of dependencies (ie. every Scala project in the workspace) whenever fetching dependencies via `wit fetch-scala` or when `wake` is determining the classpath for compilation or execution.

The startup time ends up mattering a lot. Compared to `0.11.0`, `wit fetch-scala` is a little bit faster for clean fetching, and *much* faster for invocations when there is no downloading to do (like running `wit fetch-scala` just to check everything is downloaded, or `wake` getting compilation classpaths):

For a large design, fresh installs are 17% faster with this PR, but when there isn't anything to do, it is **75%** faster (and almost all of the remaining time is due to the Scala compiler bridge fetcher, a separate problem).

0.11.0
Fresh Install: ~35s
Nothing to do: ~20s

This PR
Fresh Install: ~29s
Nothing to do: ~5s


lib/wit/scalaplugin.py
```diff
@@ -86,14 +87,52 @@ def fetch_scala_compiler_bridge(coursier, bloop_home, cache, version):
...
-    url = "http://central.maven.org/maven2/{}".format(path)
+
     print("Downloading from {}".format(url))
     urllib.request.urlretrieve(url, filename)
+
+    actual_sha256 = calc_sha256(filename)
+    if actual_sha256 != sha256:
+        suggestion = "Please delete 'scala/' directory and re-run fetch-scala!"
+        extra_info = "  Expected: {}\n  Got:      {}".format(sha256, actual_sha256)
+        msg = "Error! SHA256 mismatch for {}! {}\n{}".format(filename, suggestion, extra_info)
```
richardxia: Minor suggestion for making the code a bit easier to read: If you break out the "Error! SHA256 mismatch" part of the string into its own variable, it would allow the code to read in the same order as the message itself. At first glance I thought the error message would be printed in the `suggestion`, `extra_info`, and `msg` order, but it's actually `msg`, `suggestion`, `extra_info`, which is (correctly) the more sensible order.

jackkoenig:
I'm going to merge this after https://github.com/sifive/wit/pull/171 which will go into `0.11.1` while this is `0.12.0`

PR #175, opened 2019-11-12
Improve tests

jackkoenig:
Minor improvements:
* Make it possible to set which wit to use in regress.sh
* Purge "regress" from testing terminology

Mainly motivated by wanting to test an install from a regular wit repo. You can now run the tests by setting environment variable `WIT` and it will use that wit instead:

```
$ WIT=/path/to/wit/0.11.1 ./t/test_all.sh
```


.travis.yml
```diff
@@ -28,8 +28,8 @@ jobs:
       name: "Type Checking"
       script: mypy lib/wit/*.py
     - stage: test
-      name: "Regression Tests"
```
richardxia: Just for the record, since I'm the main one that has been asking for the terminology change: I think the phrase "[regression test](https://en.wikipedia.org/wiki/Regression_testing)" is fine, since it is refers to a test that checks that there are no regressions in behavior.

What I am against is using the word "regression" to refer to the test itself, since it means the exact opposite of what we want. To regress means to return to a previous state, not to run tests that hopefully prevent returning to a previous state.

t/test_util.sh
```diff
@@ -4,7 +4,14 @@ wit_repo='git@github.com:sifive/wit'
 test_root=$(dirname $(perl -MCwd -e "print Cwd::realpath('$0')"))
 wit_root=$(perl -MCwd -e "print Cwd::realpath('$test_root/..')")
 
-export PATH=$wit_root:${PATH}
+if [[ -z "${WIT}" ]]; then
```
richardxia: Is this `WIT` environment variable documented anywhere? Also, it's a little unintuitive that a variable named `WIT` should point to the _directory_ containing the `wit` executable, not the `wit` executable itself. Should we consider setting `WIT` to the path of the executable itself, and defaulting to `$wit_root/wit` if it's unset?

.travis.yml
jackkoenig: Hahah yeah fair enough, I'll restore it to "Regression Tests" but keep everything else the same.

t/test_util.sh
jackkoenig: To be clear, this code doesn't require that `WIT` points to the directory containing `wit`, it also works for the directory containing it, eg.
```
WIT=`which wit` ./t/test_all.sh
```
This will use the result of `which wit`.

Nothing about the tests are documented at all. I'll add a little blurb to the `README-internals`

PR #177, opened 2019-11-21
Remove unconditional directory creation prior to `git clone`.

### LINKED ISSUE #169: Error with wit add-pkg
If you `wit add-pkg package` and wit reports that the source is invalid, then if you fix it `wit add-pkg git@github.com/org/package.git`, wit will error because it already created a directory `.wit/package`

mmjconolly:
Otherwise if the clone fails we have a directory to clean up.
Instead rely on `git clone` itself to do creation and cleanup.
Closes #169

### DIFF
--- a/lib/wit/gitrepo.py
+++ b/lib/wit/gitrepo.py
@@ -79,8 +79,8 @@ class GitRepo:
             "Trying to clone and checkout into existing git repo!"
         log.info('Cloning {}...'.format(self.name))
 
-        self.path.mkdir()
-        proc = self._git_command("clone", "--no-checkout", source, str(self.path))
+        proc = self._git_command("clone", "--no-checkout", source, str(self.path),
+                                 working_dir=str(self.path.parent))
         try:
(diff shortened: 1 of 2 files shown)


lib/wit/gitrepo.py
```diff
@@ -283,11 +282,11 @@ def manifest(self, source, revision):
...
         }
 
-    def _git_command(self, *args):
+    def _git_command(self, *args, working_dir=None):
         log.debug("Executing [{}] in [{}]".format(' '.join(['git', *args]), self.path))
         proc = subprocess.run(['git', *args], stdout=subprocess.PIPE,
-                              stderr=subprocess.PIPE,
-                              cwd=str(self.path), universal_newlines=True)
+                              stderr=subprocess.PIPE, universal_newlines=True,
+                              cwd=str(self.path) if working_dir is None else str(working_dir))
```
sistillson: put each argument on it's own line
maybe move the if to it's own line

Issue #185, opened 2019-12-12
wit add-dep has confusing error message when (accidentally) run from workspace root


richardxia:
Steps to reproduce:

```sh
$ wit init test-ws -a git@github.com:sifive/wit.git
$ cd test-ws
$ wit add-dep git@github.com:sifive/wake.git
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/local/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/usr/local/lib/wit/__main__.py", line 11, in <module>
    main()
  File "/usr/local/lib/wit/main.py", line 135, in main
    add_dep(ws, args)
  File "/usr/local/lib/wit/main.py", line 227, in add_dep
    cwd_dirname = cwd.relative_to(ws.root).parts[0]
IndexError: tuple index out of range
```

jackkoenig:
😬

richardxia:
Fixed in #187.

PR #186, opened 2019-12-13
Error on add-dep in workspace root

mmjconolly:
Adds error message similar to update-dep.
Issue #185


jackkoenig:
Just a heads up that you don't need to open a new PR to change the base 🙂

PR #187, opened 2019-12-13
Error out if add-dep is run in workspace root

mmjconolly:
Adds a similar error message as update-dep.
Issue #185


lib/wit/main.py
```diff
@@ -227,8 +227,10 @@ def add_dep(ws, args) -> None:
     req_dep = dependency_from_tag(ws.root, args.pkg)
 
     cwd = Path(os.getcwd()).resolve()
-    cwd_dirname = cwd.relative_to(ws.root).parts[0]
+    if cwd == ws.root:
+        error("Cannot run add-dep from root of workspace.")
```
jackkoenig: I'm happy with this as is, but do you think it might be worth also suggesting `-C`?

richardxia: 👍 on providing instructions on how to resolve the issue. Should we also add a description of _why_ running `add-dep` from the workspace root is an error? I think the error messages are an opportunity to teach users more about correct usage of the tool as well as bits and pieces of the mental model.

jackkoenig: Yeah agreed, I'll update the error and then merge.

jackkoenig:
@richardxia what do you think of my updated messages?

lib/wit/main.py
```diff
@@ -227,8 +227,11 @@ def add_dep(ws, args) -> None:
     req_dep = dependency_from_tag(ws.root, args.pkg)
 
     cwd = Path(os.getcwd()).resolve()
-    cwd_dirname = cwd.relative_to(ws.root).parts[0]
+    if cwd == ws.root:
+        error("add-dep must be run inside of a package.\n  The cwd (which can also be set by -C) " +
```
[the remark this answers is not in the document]
richardxia: Also, you may want to just fully spell out "current working directory", as I would suspect that not everyone who is using wit may be familiar with that acronym. (Or you may confuse the [Welsh](https://en.wiktionary.org/wiki/cwd)).

lib/wit/main.py
```diff
@@ -227,8 +227,11 @@ def add_dep(ws, args) -> None:
...
+              "determines to which package a dependency will be added.")
```
richardxia: I feel the construction of this second sentence is a bit awkward, since it takes a number of detours before getting to the "main action". What if we tried flipping the order of the words like this:

> A dependency is added to the package determined by the current working directory, which can also be set by -C.

jackkoenig: That is better, updating then merging once CI passes 🙂

PR #188, opened 2019-12-14
Make any number of -v >= 4 count as SPAM

jackkoenig:
Previously, if you did too many, you ended up with INFO verbosity.
Presumably, any user putting tons of -vvvvvvv mean highest possible verbosity.


lib/wit/main.py
```diff
@@ -88,7 +88,7 @@ def main() -> None:
     subparsers.add_parser('fetch-scala', help='Fetch dependencies for Scala projects')
 
     args = parser.parse_args()
-    if args.verbose == 4:
+    if args.verbose >= 4:
         log.setLevel('SPAM')
     elif args.verbose == 3:
         log.setLevel('TRACE')
```
richardxia: I know I'm commenting on something unrelated, but what are all of these levels? These are not the standard ones defined by the Python standard library: https://docs.python.org/3/library/logging.html#levels

jackkoenig: Sam and I just defined our own. I personally think you need more granularity beyond "Debug". "Trace" is common in other languages and I believe "Spam" is also sometimes used. Totally could revise this, but sometimes it's nice to just be able to dump stuff to the screen that you can pick through.

jackkoenig: Put another way, I think "debug" is potentially useful for users, whereas "trace" and "spam" are more what one would ask a user to do then send in the bug report. Of course are two levels necessary there? Good question.

richardxia: OK, not blocking this PR, but if we have these rough ideas of what types of messages go into which level, we should write them down, because otherwise newcomers who are trying to contribute code won't know which levels are meant for which kinds of messages. I had never heard of TRACE before, but I do see that it's pretty common.

jackkoenig: Yeah agreed

PR #196, opened 2020-01-04
Add GitHub action for Wit.

richardxia:
Per https://github.com/sifive/api-scala-sifive/pull/21#issuecomment-570737983, here is a GitHub action that is included as part of the Wit repository itself. See the PR itself for more details, but I have a working example of using this in https://github.com/sifive/api-scala-sifive/pull/21


actions/wit/README.md
```diff
@@ -0,0 +1,22 @@
...
+## Example Usage
+
+```yaml
+- name: 'Wit Init'
+  uses: sifive/wit/actions/wit@v0.11.2
+  with:
+    command: init
+    arguments: |
+      workspace_dir
+      -a ./local-packages/api-scala-sifive
```
jackkoenig: The `local-packages` thing is a product of where you tell the Github Actions CI to put the repo right?

richardxia: Yeah, maybe I should reword this to `-a ./path/to/local/git/repo` to make it clearer that `local-packages` is not a special name.

jackkoenig:
Will you back port this to `0.11.x` and crack `0.11.2`?

richardxia:
> Will you back port this to `0.11.x` and crack `0.11.2`?

Talked offline, but it sounds like we're going to crack an `0.12.0` instead.

I'm going to leave the existing version numbers in both the Dockerfile and the README for the action, but @mmjconolly will take care of creating the release commit that bumps them. I think it'd be better for us to make the version number change in the release commit, since that will more closely match how we'll do this in the future anyway.

PR #206, opened 2020-02-07
Pip install?

robertd-sifive:
Add a setup.py file to make it so you can pip install wit and use it directly from python. It will also add a wit executable to your local python environment.

Had to nest the setup.py within lib though to keep the existing directory structure, so I added docs on how to refer to it from pip or equivalently in requirements.txt.


mmjconolly:
Thanks for this,

Could you talk a little about the following points
- The `README.md` moved to `lib` (also please use `git mv` when doing so, makes the diff cleaner)
- `./lib/bin/wit` is implemented differently from `./wit`

robertd-sifive:
Hrm, tried to git mv, but I think because I created the symlink it blew out the changes. These are the changes to the README:

* Typo fix:
    - before: `Wit is a tool for managing workspaces...`
    - after: `Wake is a tool for managing workspaces...`
* Pip explanation:
---------------------------------------------
### Installation with pip
Wit is not currently registered with [pypi](https://pypi.org), however it can be installed directly from Github using pip:

```bash
pip install -U git+https://github.com/sifive/wit@master#egg=wit&subdirectory=lib
```

This will also place a `wit` executable in your python (virtual) environment’s `bin` directory, which you may need to add to your shell’s $PATH.

This url scheme can be used to point at a specific commit hash. Note that the source code is located in `./lib`, so the subdirectory must be specified in the url. Showing each part of the url more explicitly:

```bash
package_name='wit
commit_hash='master'
source_url='https://github.com/sifive'
source_name='https://github.com/sifive/wit'
relative_src_dir='lib'
pip install -U 'git+'"$source_url"'@'"$commit_hash"'#egg='"$package_name"'&subdirectory='"$relative_source_dir"
```
----------------------------

I feel like there is a way for setup.py to point to a package that is not immediately adjacent to it, but I couldn't figure it out from the docs, but I've used this trick in other repos where I've needed to nest python packages.

I went ahead and added the exec to my version of the wit wrapper script in ./lib/bin/ to give it the same no-forking behavior, but since it's going to be installed into a virtual environment, the PYTHONPATH shouldn't be messed with there since PYTHONPATH should point at the virtualenv. However, calling from the top like in `./wit`, ensuring the right PYTHONPATH makes sense I think.

lib/README.md
```diff
@@ -0,0 +1,169 @@
...
+Each package may optionally contain a wit-manifest.json file which defines other packages upon which it depends.
+Wit resolves this hierarchy of dependencies and generates a flattened directory structure in which each package
+may exist only once.
+
+## How does Wit deduplicate packages?
+When multiple versions of the same package are requested, Wit chooses the latest requested version, making sure the selected version's commit is a descendant of every other requested versions' commits.
+
+## Getting started
+
+The best way to learn wit is to check out the [tutorial](share/doc/wit/tutorial.md).
```
richardxia: Note that every single one of these relative URLs is no longer correct, at least from the perspective of the `lib/README.md` location.

I'm not sure we can really get away with using the same README for both GitHub and the Python package, given that the README is written from the perspective of the source repository, not of the Python package itself. I know it's a copout, but if you must have a README in the Python package, can you just create a separate document that  links to the GitHub repository?

lib/setup.py
```diff
@@ -0,0 +1,29 @@
...
+    "scala-bridge-fetcher_2.12-0.1.0.jar"
+]
+
+scripts = [
+    'bin/wit',
+]
+
+setup(
+    name=package_name,
+    version='0.1',
```
richardxia: Let's make this match the actual version number of the tool. At the moment, we are at 0.12.0.

lib/setup.py
```diff
@@ -0,0 +1,29 @@
...
+assets = [
+    "scala-bridge-fetcher_2.12-0.1.0.jar"
+]
+
+scripts = [
+    'bin/wit',
+]
+
+setup(
+    name=package_name,
```
richardxia: Per https://packaging.python.org/guides/distributing-packages-using-setuptools/#name, this is supposed to be the name of the package as listed on PyPI. Given that PyPI already has a [separate package by the name of `wit`](https://pypi.org/project/wit/), I don't think we should name this `wit`, since it could confuse any tools that assume that the value of this field can be used to query PyPI for the package.

Note that the name field here is not the same thing as the name that you use in `import` statements, since hyphens and periods are explicitly allowed, despite not being valid in Python identifiers. Also, for a real-life example of this, see the [PyYAML library](https://github.com/yaml/pyyaml/blob/2f463cf5b0e98a52bc20e348d1e69761bf263b86/setup.py#L2), which has a PyPI name of `PyYAML` but an import package name of `yaml`.

If no one has any better suggestions, I will suggest a package name of `wit-sifive`.

richardxia:
> The README.md moved to lib (also please use git mv when doing so, makes the diff cleaner)

`git mv` isn't actually a first class operation within the git internals, and it's actually no different than `git rm` and `git add`; detection of a move is done post hoc and heuristically by diffing tools, and because you've modified the README enough, GitHub isn't detecting it as a move. See https://git.wiki.kernel.org/index.php/GitFaq#Why_does_Git_not_.22track.22_renames.3F if you care about the details.

lib/README.md
```diff
@@ -0,0 +1,169 @@
...
+```
+
+The Makefile will create a directory with the version (it even works for commits between tags)
+and copy the contents of contents of the local clone excluding the tests and metadata.
+
+### Installation with pip
+Wit is not currently registered with [pypi](https://pypi.org), however it can be installed directly from Github using pip:
+
+```bash
+pip install -U git+https://github.com/sifive/wit@master#egg=wit&subdirectory=lib
```
richardxia: Please don't encourage people to install from `master`. We've put in effort to actually do versioned releases of Wit, so I'd much rather people install from a tagged version.

If the name [`wit` were not already taken on PyPI](https://pypi.org/project/wit/), I'd also just recommend that we publish directly to PyPI. We might want to register a name like `wit-sifive` on PyPI so that people can install from there instead of pointing directly to GitHub.

lib/setup.py
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env python3
+
+from distutils.core import setup
+from pathlib import Path
+
+package_name = 'wit'
+
+assets = [
+    "scala-bridge-fetcher_2.12-0.1.0.jar"
```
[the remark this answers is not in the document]
jackkoenig: Yeah, I would wait on merging this until we've just removed the Scala plugin, should simplify a lot of things

lib/README.md
mmjconolly: I also like the publish-to-PyPI, we could then remove the current `make install`

lib/setup.py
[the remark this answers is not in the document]
mmjconolly: @robertd-sifive Jack has removed the Scala plugin from Wit now

richardxia:
Closing, since #220 has done this and is now merged in.

PR #207, opened 2020-02-14
Add how to guide on wit inspect

richardxia:
Addresses a few things in https://github.com/sifive/wit/issues/200.

This PR does two main things:

1. Move parts of the README into a new `docs/how-to-guides.adoc` file (https://github.com/sifive/wit/commit/0bdfe8b2a54a1ed3f3e72e4a8993ba0230b6b282)
2. Document the `wit inspect` command (https://github.com/sifive/wit/commit/a52b56d53ac4a7b5912614f0d8a4a01fe5003996)

PR #208, opened 2020-02-14
Add support for GitHub access tokens

nategraff-sifive:
Accept the `token` parameter for authenticating to GitHub via access token.


nategraff-sifive:
Just confirmed that `https://x-access-token:$TOKEN@github.com` works for accessing a private repo.

actions/wit/entrypoint.sh
```diff
@@ -5,7 +5,14 @@ set -euxo pipefail
 : ${INPUT_COMMAND:?'The `command` input argument must be set'}
 
 if [[ -n "$INPUT_FORCE_GITHUB_HTTPS" ]]; then
-  git config --global url.'https://github.com/'.insteadOf 'git@github.com:'
+  if [[ -n "$INPUT_TOKEN" ]]; then
+    # Don't echo the token
+    set +x
+    git config --global url."https://x-access-token:${INPUT_TOKEN}@github.com/".insteadOf 'git@github.com:'
```
mmjconolly: I realize you didn't introduce this usage of `--global`, but I would like to make the comment that this command would change the behavior of git commands unrelated to the current workspace

nategraff-sifive: I guess the question is: Does removing `--global` break anything about the intended usage of the action?

nategraff-sifive: [The `git-config` documentation](https://git-scm.com/docs/git-config#Documentation/git-config.txt---global) seems to say that without `--global`, the changes only apply to the local repository's `.git/config`, which doesn't actually work here because we want these changes to apply to every repository in the workspace.

nategraff-sifive: Since these commands are being run within the context of the wit container, I think the config file being modified is the one inside the container. Therefore the scope actually is limited to the workspace because the container only persists for the duration of this GitHub action.

nategraff-sifive:
Hold up, I think I have a better way. This way leaves the auth token in the git remote url, but looking at the source for actions/checkout there's a way to set the token as an extra header in the git config so that it never leaves the container.

actions/wit/entrypoint.sh
mmjconolly: Oh I misread the whole diff, I didn't realize this was in a github action

nategraff-sifive:
Okay, so setting http.extraheader is way too fragile. I'm going back to the URL version.

I was also wrong about this persisting in the remote urls. Since this rewrites URLs the repository remote urls never update and the access tokens are still isolated to the container, so there's no benefit to the extraheader method.

actions/wit/entrypoint.sh
```diff
@@ -5,7 +5,14 @@ set -euxo pipefail
 : ${INPUT_COMMAND:?'The `command` input argument must be set'}
 
 if [[ -n "$INPUT_FORCE_GITHUB_HTTPS" ]]; then
-  git config --global url.'https://github.com/'.insteadOf 'git@github.com:'
+  # Don't echo the token
+  set +x
+  if [[ -n "$INPUT_TOKEN" ]]; then
+    git config --global url."https://x-access-token:${INPUT_TOKEN}@github.com/".insteadOf 'git@github.com:'
```
richardxia: Can you double check that other regular git/wit operations don't print out the access token value?

BTW, nice job on investigating how we can authenticate against private repos like this!

richardxia:
Also, I made the exact same mistake in https://github.com/sifive/wit/pull/207, but you should probably switch the base branch from `0.12.x` to `master`.

actions/wit/entrypoint.sh
nategraff-sifive: In my experiments I didn't see it leak. It looks to me that the git porcelain doesn't see the rewritten URL.

nategraff-sifive:
Switched the base branch to master and rebased.

actions/wit/entrypoint.sh
richardxia: After talking to @jackkoenig offline, I realized that maybe we should make this GitHub Action parameter more generic. The underlying authentication method you're using is just [HTTP Basic Authentication](https://en.wikipedia.org/wiki/Basic_access_authentication), and the `x-access-token` is just a special username. Should we replace this with more generic, HTTP basic auth parameters?

Alternatively, should we give this parameter a much more specific name, such as including `GITHUB` in the name itself, so that it's clear that it's only for use with GitHub?

nategraff-sifive: I agree on adding a username field, I know Azure uses a similar mechanism. How about:
`http_auth_username` (with a default value of "x-access-token") and `http_auth_token`?

richardxia:
@mmjconolly, @jackkoenig, it looks like when a PR is submitted from a fork, the `branch` Travis CI build is not executed. I'm guessing this is for security purposes, since the branch technically lives in a fork. I'm going to change the repo settings to remove the requirement that the Travis CI Branch build succeed before being able to merge a PR, since I think we only need the PR build to succeed. Let me know if you don't think I should do this, and we can revert the change.

actions/wit/entrypoint.sh
richardxia: LGTM!

Issue #210, opened 2020-02-24
wit init --from-lockfile


mmjconolly:
It could be nice to have the ability to seed a workspace from a previous `wit-lock.json` from another workspace.
This could be used for
- reproducing an issue reported
- moving between machines and wants to re-init with the same revisions
- archival purposes

PR #211, opened 2020-02-25
restore command recreates workspace from lockfile

mmjconolly:
A `wit restore` will populate a fresh workspace using
provided wit-workspace.json and wit-lock.json files.
No dependency resolution happens.

Resolves https://github.com/sifive/wit/issues/210

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,24 @@ and copy the contents of contents of the local clone excluding the tests and met
 See the [How To Guides](doc/how-to-guides.adoc) for list of guides for common wit operations.
 
 
+### Restore a previous workspace
+
+If you have a matching pair `wit-lock.json` and `wit-workspace.json` from another workspace, you can create
+a new clean workspace by pointing at the previous workspace via the `restore` sub-command.
+
+    $ wit restore -n <new-workspace-name> -w $OTHER_WS
+
+Dependencies are only specified by the `wit-lock.json` file, no other dependency resolution is performed.
+If you are keeping your wit-lock.json and wit-workspace.json under revision control you can do an in-place
+restore
+
+    $ git clone http://example.com/my_workspaces
+    $ cd my_workspaces/project1
+    $ ls
+    wit-lock.json wit-workspace.json
+    $ wit restore
+
+
 ## Autocompletion
 
 Tab completion can be enabled via `source complete.bash`. If you want it to persist, see below.
--- a/lib/wit/main.py
+++ b/lib/wit/main.py
@@ -12,6 +12,7 @@
 
 import subprocess
 import sys
+import shutil
 import os
 from .witlogger import getLogger
 from .workspace import WorkSpace, PackageNotInWorkspaceError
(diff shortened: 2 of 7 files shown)


mmjconolly:
Besides the speed gain of not having to perform the dependency resolution and being able to directly parallel clone, I'm thinking that this `restore` command also provides a nice experience for those that might want to revision control their lockfiles.

```
$ git clone git@github.com/some-name/workspaces workspaces
Cloning into 'workspaces'...
done.

$ cd workspaces/

$ tree
.
|-- project1
|   |-- wit-lock.json
|   `-- wit-workspace.json
`-- project2
    |-- wit-lock.json
    `-- wit-workspace.json

$ cd project2

$ wit restore
Cloning repo1...
Cloning repo2...
Checking out 'repo1' at 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa'
Checking out 'repo2' at 'bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb'

$ wit update-pkg repo1::cccccccccccccccccccccccccccccccccccccccc

$ git diff
diff --git a/project2/wit-workspace.json b/project2/wit-workspace.json
index aaaaaa..bbbbbb 100644
--- a/project2/wit-workspace.json
+++ b/project2/wit-workspace.json
@@ -1,6 +1,6 @@
 [
     {
-        "commit": "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
+        "commit": "cccccccccccccccccccccccccccccccccccccccc",
         "name": "repo1",
         "source": "git@github.com:user/repo1.git"
     },
```

README.md
```diff
@@ -77,6 +77,24 @@ You can always checkout remote branches by specifying the remote as well
...
+If you are keeping your wit-lock.json and wit-workspace.json under revision control you can do an in-place
+restore
+
+    $ git clone http://example.com/my_workspaces
+    $ cd my_workspaces/project1
+    $ ls
+    wit-lock.json wit-workspace.json
+    $ wit restore
+
+
```
jackkoenig: Rather than having two optional but related arguments `-l` and `-w`, what if instead you just point to a directory that is assumed to contain both?
```
wit restore <new-workspace-name> [--from-workspace|-w] </path/to/workspace>
```
This seems more consistent with the second form requiring no arguments.

mmjconolly: I guess my feeling was that no one would use the `-w` and `-l` much. I can change it to the directory style you've suggested.

On a somewhat related note, I feel like they should just be 2 sections in the one file

mmjconolly:
I've removed the --force option, it's usage was a little weird on already created workspaces (for example, pushing all the HEADs to what the lock says, but not dealing with unclean repos).
So instead it's easier to just start by requiring that the workspace is 'new' and if we need a --force we can try figuring it out properly later

README.md
mmjconolly: That `--from-workspace` change is made now

jackkoenig:
@mmjconolly I'll leave this for you to merge so you can decide if you want to squash or merge

mmjconolly:
Needed a small fixup due to the new `jobs` param on `Workspace`

PR #213, opened 2020-02-29
Add a 'foreach' command

mmjconolly:
This is inspired by `git submodule foreach` [[link](https://git-scm.com/docs/git-submodule#Documentation/git-submodule.txt-foreach--recursiveltcommandgt)]

```
$ wit foreach --help
usage: wit foreach [-h] [--quiet] [--continue-on-fail] cmd ...

Perform a command in each repository directory.
The repository list is created by reading records contained in 'wit-lock.json'.

If used, options --quiet and --continue-on-fail must be specified before the command.

The following environment variables are exposed to the command:
    WIT_REPO_NAME    repository name
    WIT_REPO_PATH    path to repository
    WIT_LOCK_SOURCE  inital source location of repository
    WIT_LOCK_COMMIT  commit recorded in lockfile, actual repository contents could be different
    WIT_WORKSPACE    path to root of wit workspace

positional arguments:
  cmd                 command to run in each repository
  args                arguments for the command

optional arguments:
  -h, --help          show this help message and exit
  --quiet             Supress printed lines indicating working repository
  --continue-on-fail  run the command in each repository regardless of
                      failures
```

```
$ wit foreach 'echo $WIT_REPO_NAME; git rev-parse HEAD'
Entering 'repo1'
repo1
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
Entering 'repo2'
repo2
bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
```

### DIFF
--- a/lib/wit/main.py
+++ b/lib/wit/main.py
@@ -92,6 +92,9 @@ def main() -> None:
             elif args.command == 'update':
                 update(ws, args)
 
+            elif args.command == 'foreach':
+                foreach(ws, args)
+
             elif args.command == 'fetch-scala':
                 fetch_scala(ws, args, agg=False, jar=args.jar)
 
@@ -106,6 +109,34 @@ def main() -> None:
         error(e)
     except AssertionError as e:
         raise WitBug(e)
+
+
+def foreach(ws, args):
+    has_fail = False
+    for pkg in ws.lock.packages:
+        env = os.environ.copy()
+        env["WIT_REPO_NAME"] = pkg.name
+        env["WIT_REPO_PATH"] = str(ws.root / pkg.name)
+        env["WIT_LOCK_SOURCE"] = pkg.source
+        env["WIT_LOCK_COMMIT"] = pkg.revision
+        env["WIT_WORKSPACE"] = str(ws.root)
+
+        log.info("Entering '{}'".format(pkg.name))
+
+        location = str(ws.root / pkg.name)
+        proc = subprocess.run([args.cmd, *args.args], shell=True,
+                              env=env, cwd=location, universal_newlines=True)
+
+        if proc.returncode != 0:
+            has_fail = True
+            command = " ".join([args.cmd, *args.args])
+            log.error("Command '{}' in '{}' failed with exitcode: {}"
+                      .format(command, location, proc.returncode))
+            if not args.continue_on_fail:
+                sys.exit(proc.returncode)
+
+    if has_fail:
+        sys.exit(1)
 
 
 def parse_repo_path(args):
--- a/lib/wit/parser.py
+++ b/lib/wit/parser.py
@@ -99,6 +99,31 @@ inspect_group = inspect_parser.add_mutually_exclusive_group()
 inspect_group.add_argument('--tree', action="store_true")
 inspect_group.add_argument('--dot', action="store_true")
 
+# ********** foreach subparser **********
+foreach_parser = subparsers.add_parser(
+    'foreach',
+    help='perform a command in each repository directory',
+    formatter_class=argparse.RawDescriptionHelpFormatter,
+    description="""
+Perform a command in each repository directory.
+The repository list is created by reading records contained in 'wit-lock.json'.
+
+Any options, such as --continue-on-fail must be specified before the command.
+
+Wit sets the following environment variables for each invocation of the command:
+    WIT_REPO_NAME    repository name
+    WIT_REPO_PATH    path to repository
+    WIT_LOCK_SOURCE  initial source location of repository
+    WIT_LOCK_COMMIT  commit recorded in lockfile, actual repository contents could be different
+    WIT_WORKSPACE    path to root of wit workspace""")
+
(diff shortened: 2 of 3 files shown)


lib/wit/main.py
```diff
@@ -108,6 +111,37 @@ def main() -> None:
         raise WitBug(e)
 
 
+def foreach(ws, args):
+    packages = sorted(ws.lock.packages, key=lambda x: str.lower(x.name))
```
richardxia:
```suggestion
    packages = sorted(ws.lock.packages, key=lambda x: x.name.lower())
```

I think it's a bit more idiomatic in Python to call this in method form rather than function form. See https://docs.python.org/3.5/library/stdtypes.html?highlight=lower#str.lower

lib/wit/main.py
```diff
@@ -108,6 +111,37 @@ def main() -> None:
...
+
+    has_fail = False
+    for pkg in packages:
+        env = os.environ.copy()
+        env["WIT_REPO_NAME"] = pkg.name
+        env["WIT_REPO_PATH"] = str(ws.root / pkg.name)
```
richardxia: Is this the same as `pkg.repo.path`?

lib/wit/main.py
jackkoenig: Rather than sorting, I think it should be left in the order read from the lock file. I thought that in `lock.py` we already lock the packages into an `OrderedDict`, but it looks like we only write them that way. My preference would be to make the ordering deterministic that way instead of via sorting here.

lib/wit/parser.py
```diff
@@ -99,6 +99,33 @@ def err(msg):
...
+
+foreach_parser.add_argument('--quiet', action='store_true',
+                            help='Supress printed lines indicating working repository')
+foreach_parser.add_argument('--continue-on-fail', action='store_true',
+                            help='run the command in each repository regardless of failures')
+
+# 'cmd' and 'args' eventually become one list, but this forces at least one input string
+foreach_parser.add_argument('cmd', help='command to run in each repository')
+foreach_parser.add_argument('args', nargs=argparse.REMAINDER, help='arguments for the command')
+
```
jackkoenig: Great job on the detailed description.

One minor clarification is that I was a little confused by what the environment variables mean when reading it. I would rephrase slightly to make it more clear that these are environment variables set by `wit` for each `command` rather than ones the user can set. I'm not sure on the best phrasing but here's a try:
```
Wit sets the following environment variables for each invocation of the command:
    WIT_REPO_NAME    repository name
    ...
```

lib/wit/main.py
```diff
@@ -108,6 +111,37 @@ def main() -> None:
...
+        env["WIT_LOCK_SOURCE"] = pkg.source
+        env["WIT_LOCK_COMMIT"] = pkg.revision
+        env["WIT_WORKSPACE"] = str(ws.root)
+
+        if not args.quiet:
+            print("Entering '{}'".format(pkg.name))
```
jackkoenig: Should this be governed by the logger? I'm wondering if the whole `--quiet` option should be folded into that then this print would just be `log.info`.

lib/wit/main.py
mmjconolly: Just tried it, it's value is `None`, I'm guessing we never create the repo object

jackkoenig: Yeah this is one of the weird parts of the implementation, I *think* `pkg.repo` is set by running `resolve` or something like that. We should probably set it for the `Package`s in the Lockfile though when we construct the `Workspace` object.

lib/wit/main.py
mmjconolly: I'm trying to copy the usage from `git submodule foreach` usage, 
from the git docs:
```Unless given --quiet, foreach prints the name of each submodule before evaluating the command. ```

jackkoenig: I agree with generally following git's lead, but since we have unified logging (although I think it currently lacks a mechanism to _lower_ the verbosity which we obviously need), I would think it would be better to keep it unified with the logging API. I think it would be strange if just 1 command ignores your logging settings.

lib/wit/main.py
mmjconolly: Happy to loop in the order they're in the lock file

lib/wit/main.py
mmjconolly: If we make a global `--quiet`, we'd need to do it properly and make all the `print()` in the codebase conditional rather than just the log lines.

How about if I make the print line in this snippet unconditional, but print the line to stderr so that people who don't want it can filter it out with 2>/dev/null ?

mmjconolly: though grep-ing, we don't actually have that many `print()`

mmjconolly: I'll break this stuff into a separate diff

mmjconolly:
Not sure how I closed this

lib/wit/main.py
mmjconolly: https://github.com/sifive/wit/pull/214

PR #216, opened 2020-03-25
Add log line to read_manifest

mmjconolly:
If someone has a manually modified manifest file, they otherwise would get a stacktrace if it's not valid json.

### DIFF
--- a/lib/wit/manifest.py
+++ b/lib/wit/manifest.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 
 import json
+import sys
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


jackkoenig:
It's failing lint because it expected tabbing to be a multiple of 4

mmjconolly:
I'm thinking I should go put a wrapper around all the `json.loads` and related functions, rather than just this one

jackkoenig:
> I'm thinking I should go put a wrapper around all the json.loads and related functions, rather than just this one

That would probably be best

PR #217, opened 2020-04-01
Remove the Scala plugin

jackkoenig:
This has been moved to api-scala-sifive: https://github.com/sifive/api-scala-sifive/pull/27


mmjconolly:
nit: `complete.bash` has some references to scala plugin

jackkoenig:
Nice catch, fixed

jackkoenig:
CI is soooo much faster now 🙂

richardxia:
> CI is soooo much faster now 🙂

It looks like the slowest part is now building the Dockerfile for the wit action. I bet I could make this faster by pulling down the previously built image and then building on top of that :P.

PR #218, opened 2020-04-02
Add environment variable for git clone --reference

mmjconolly:
Point at another workspace as a seed, to speed up `wit init`, especially on slower networks. Not all the required repositories have to be in the seed workspace. 

Example usage 
```
$ export WIT_WORKSPACE_REFERENCE=$HOME/other-workspace
$ wit init new-workspace -a git@github.com:acme/foo
```

I'm not sure that an environment variable is the best way to go here, but it looked like there were a lot of paths to follow if we were to thread the variable through from main.
Let me know if you think I should do it a different way.

The commit message:
```
git clone's --reference[-if-able] option uses a local checkout of a
repository to seed the objects that would be fetched during a clone.
On the same filesystem this will result in hardlinks, and across
filesystems it uses symlinks. Using symlinks is an issue if the original
repository is removed, it will break the second repository.
To ensure the operation is safe, '--dissociate' is added to git clone
to instead copy the objects. This is likely still faster than network.

The environment varible is WIT_WORKSPACE_REFERENCE which should be a
directory containing repositories you would like to use as a seed,
such as another wit workspace.
Any repositories that are missing are cloned from the network as before.
```


mmjconolly:
On a large workspace on my home network, it's cuts the time in half

lib/wit/gitrepo.py
```diff
@@ -79,8 +80,14 @@ def clone(self, source, name):
             "Trying to clone and checkout into existing git repo!"
         log.info('Cloning {}...'.format(self.name))
 
-        proc = self._git_command("clone", "--no-checkout", source, str(self.path),
-                                 working_dir=str(self.path.parent))
+        cmd = ["clone"]
+        env_ref_path = os.getenv("WIT_WORKSPACE_REFERENCE")
```
richardxia, line 84: I  am in general usually a bit reluctant to read environment variables in different "leaf-level" methods within a large program, since it's pretty easy to lose track of them. Even if we do go with environment variables for the user-facing interface, I'd probably prefer centralizing the environment variable in a single module, perhaps adjacent to the command-line argument parsing. I am not familiar, however, enough with Wit to give you any suggestions on how to best thread this data through the Wit code, so I'd probably need @jackkoenig to comment here.

Now, putting that question aside, I think an environment variable as the user interface makes sense here, since it's harmless if it has a bad value (with `-if-able`), and it's kind of nice to set it as a globally exported shell variable so you can use it in multiple places without changing your wit commands. That said, I don't have a strong opinion either way, and we could even do both (CLI option should win, in that case).

mmjconolly:
Added emulation of the `if-able` portion of `--reference-if-able`

lib/wit/gitrepo.py
```diff
@@ -89,6 +95,17 @@ def clone(self, source, name):
             else:
                 raise
 
+    # Only newer git versions can use '--reference-if-able', so we emulate the 'if-able'.
+    def git_reference_capable(self):
+        env_ref_path = os.getenv("WIT_WORKSPACE_REFERENCE")
```
richardxia, line 100: I still really think we should do the environment handling in one place, rather than allowing any portion of the program grab at environment variables.

Also, you're missing a check to see whether this actually returned a None value or not, which means just below you are casting a `None` to a `Path`.

lib/wit/gitrepo.py
```diff
@@ -89,6 +95,17 @@ def clone(self, source, name):
...
+        paths = [Path(env_ref_path) / self.name,
+                 Path(env_ref_path) / (self.name+'.git')]
+        for path in paths:
+            log.warn("{} {}".format(path, os.path.isdir(path)))
+            if os.path.isdir(path):
+                return True, ["--reference", str(path), "--dissociate"]
+        return False, []
```
richardxia, line 107: This looks like a Go-ism. In Python, you would just return either an empty or non-empty list, and you can just put the returned list directly in the `if` statement, since empty lists are falsey and non-empty lists are truthy.

In addition, in this specific case you could actually just `.extend()` the return values whether or not it's empty since the `.extend()` method is a no-op if you're extending with an empty list.

lib/wit/gitrepo.py
```diff
@@ -89,6 +95,17 @@ def clone(self, source, name):
             else:
                 raise
 
+    # Only newer git versions can use '--reference-if-able', so we emulate the 'if-able'.
+    def git_reference_capable(self):
+        env_ref_path = os.getenv("WIT_WORKSPACE_REFERENCE")
+        paths = [Path(env_ref_path) / self.name,
+                 Path(env_ref_path) / (self.name+'.git')]
+        for path in paths:
+            log.warn("{} {}".format(path, os.path.isdir(path)))
```
richardxia, line 104: This isn't really a warning, right, since there's nothing unusual or unexpected about this codepath? I think you want this to be DEBUG, SPAM, or TRACE (I always forget the levels that we've defined here).

lib/wit/gitrepo.py
```diff
@@ -89,6 +95,17 @@ def clone(self, source, name):
             else:
                 raise
 
+    # Only newer git versions can use '--reference-if-able', so we emulate the 'if-able'.
+    def git_reference_capable(self):
```
richardxia, line 99: I think we should 1) give this method a name with a leading underscore to indicate that it is "private" and 2) we should probably give it a name that gives a better hint at what its return value is. "git reference capable" to me sounds like an adjective, and I would expect this to return a boolean value, but you're actually returning a list of command line arguments to append to a `clone` operation.

Since this is a non-mutating method with no arguments and that returns a non-boolean value, I'd probably name it something like `def _git_reference_options()`

lib/wit/gitrepo.py
```diff
@@ -89,6 +95,17 @@ def clone(self, source, name):
...
+        env_ref_path = os.getenv("WIT_WORKSPACE_REFERENCE")
+        paths = [Path(env_ref_path) / self.name,
+                 Path(env_ref_path) / (self.name+'.git')]
+        for path in paths:
+            log.warn("{} {}".format(path, os.path.isdir(path)))
+            if os.path.isdir(path):
```
richardxia, line 105: Since these are already `Path` objects, you should probably be using the Python 3.4-style [`path.is_dir()`](https://docs.python.org/3/library/pathlib.html#pathlib.Path.is_dir). See https://docs.python.org/3/library/pathlib.html#correspondence-to-tools-in-the-os-module for a table that maps from the older `os.path` functions to the newer `Path` methods.

lib/wit/gitrepo.py
mmjconolly: It's not a warning, should have removed entirely

jackkoenig:
Also nice documentation, thanks :)

One thought is maybe this variable should be mentioned in `--help` somehow; perhaps a longer description like the one you did for the `foreach` command (https://github.com/sifive/wit/blob/master/lib/wit/parser.py#L104) but for the top-level parser?

lib/wit/gitrepo.py
```diff
@@ -89,6 +90,19 @@ def clone(self, source, name):
             else:
                 raise
 
+    # Use git clone's '--reference' to point at a local repository cache to copy objects/commits
+    # to save network traffic. Any missing objects/commits are downloaded from the true remote.
+    # Only newer git versions can use '--reference-if-able', so we emulate the 'if-able' bit.
```
richardxia, line 95: Python has first-class syntax support for documentation strings, which should be placed immediately below the function `def` line as a string (usually a multiline string with `"""`). We should follow the common Python convention, since it's pretty ubiquitous and tools for generating HTML docs from code and even IDEs will read from it.

See https://www.python.org/dev/peps/pep-0257/ for the spec and for a few examples.

share/doc/wit/tutorial.md
```diff
@@ -268,3 +268,35 @@ wit status
 
 There we have it! We have made package `foo` depend on package `bar`.
 
+
+## Using a local repository cache for a faster 'wit init'
```
richardxia, line 272: Nice docs!

I think we may want this to live somewhere besides the tutorial, since I wouldn't consider it "essential" enough for an introduction tutorial to Wit. I had created a ["how-to guides"](https://github.com/sifive/wit/blob/261f9c3af98302aa2f58b3fbed66586047457273/docs/how-to-guides.adoc) document, which is an unordered collection of useful recipes, although I now realize is probably in the wrong directory (`wit/docs/` instead of `wit/share/doc/wit`).

How about we move these docs there, and then possibly move the how-to-guides into the `share/` directory, either in this PR or in a followup PR?

mmjconolly: I kinda think a top level docs directory is way more discoverable than share/doc

mmjconolly: moved it all into `Documentation/`

jackkoenig: `/share/doc/wit` was following the Debian standard. I don't really care about following that, but if we're going to move it the top, I think `doc/` is more common and consistent with the other directory names (eg. `lib`)

PR #220, opened 2020-04-09
setup.py for pip install

mmjconolly:
Add a `setup.py` so that we can `pip install`.
This has an earlier PR https://github.com/sifive/wit/pull/206 by @robertd-sifive that I leaned on heavily as I'm not very familiar with python packaging, but I had some enthusiasm to try get the ball rolling on this again.

The 0.12.0 version is just a placeholder version until we decided to actually publish to PyPi

### DIFF
--- a/lib/wit/main.py
+++ b/lib/wit/main.py
@@ -24,6 +24,7 @@ from .gitrepo import GitRepo, GitCommitNotFound
 from .manifest import Manifest
 from .package import WitBug
 from .parser import parser, add_dep_parser
+from .version import __version__
 import re
 
 log = getLogger()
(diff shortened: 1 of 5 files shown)


lib/wit/main.py
```diff
@@ -25,6 +25,7 @@
 from .package import WitBug
 from .parser import parser, add_dep_parser
 import re
+from pkg_resources import get_distribution, DistributionNotFound
```
richardxia, line 28: I think you don't want to import this stuff in your actual Python program itself, since it causes you to depend on setuptools. Python packaging is kind of a mess and depending directly on setuptools is a pretty good way of running into trouble, so we should try to keep it out of the main app and only in the packaging scripts, if at all.

@robertd-sifive, @nategraff-sifive, do you have any advice here?

mmjconolly: Mostly just looking for a way to access the version specified from setup.py, happy to try another way

mmjconolly: Just found https://packaging.python.org/guides/single-sourcing-package-version/

richardxia: Looks like there are a bunch of techniques in https://packaging.python.org/guides/single-sourcing-package-version/. It's been a while since I've done this, but I think there's no one way that is "great", so we just just pick the least evil one.

mmjconolly: I've removed the __version__ file that was created by the Makefile, created a lib/wit/version.py

mmjconolly:
When should we do a `0.13` release, can try pushing to PyPI and writing a 'how to release' doc in the one shot

PR #221, opened 2020-04-13
Error on duplicate repo entries in manifest

mmjconolly:
While looking at how to make `wit init` more parallel, I found a repo that has duplicate entries for a single repository at different commits: https://github.com/sifive/freedom-e-sdk/pull/463

This diff make it an error to do so.

### DIFF
--- a/lib/wit/manifest.py
+++ b/lib/wit/manifest.py
@@ -64,8 +64,8 @@ class Manifest:
         # Fail if there are dependencies with the same name
         dep_names = [dep['name'] for dep in json_content]
         if len(dep_names) != len(set(dep_names)):
-            dup = [x for x in dep_names if dep_names.count(x) > 1][0]
-            raise Exception("Two dependencies have the same name in {}: {}".format(location, dup))
+            dup = set([x for x in dep_names if dep_names.count(x) > 1])
+            raise Exception("Two dependencies have the same name in '{}': {}".format(location, dup))
 
         # import here to prevent circular dependency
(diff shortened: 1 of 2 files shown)


richardxia:
My only concern with this is that if some deeply nested repo has an invalid wit manifest file, you as a higher-level repo will not be able to do anything to deal with it, short of changing the low-level repo, which you may not have easy write access to.

Was freedom-e-sdk the only repo that you found with a duplicated manifest entry? If that's the case, then maybe this is fine, especially since @jackkoenig's recent changes to the Wit Scala plugin will also be a backwards-incompatible change anyway.

lib/wit/manifest.py
```diff
@@ -52,18 +52,21 @@ def write(self, path):
...
+        return Manifest.process_manifest(content, path)
 
     @staticmethod
-    def process_manifest(json_content):
+    def process_manifest(json_content, location):
+        # Fail if there are dependencies with the same name
+        dep_names = [dep['name'] for dep in json_content]
+        if len(dep_names) != len(set(dep_names)):
+            dup = [x for x in dep_names if dep_names.count(x) > 1][0]
+            raise Exception("Two dependencies have the same name in {}: {}".format(location, dup))
```
richardxia, line 68: Since you're already counting up all the duplicates, any reason not to list all of them in the error message? It's probably slightly less annoying to user who actually encounters this case to get the full list of all duplicates rather than having to fix them one by one.

PR #222, opened 2020-04-13
Partially parallel 'wit init'

mmjconolly:
This adds some amount of parallelism to the `wit init`'s git clone process.

When a package (or dependency) has multiple dependencies, those dependencies are now downloaded/cloned as a batch in parallel.
Going down one level you will create more batches of parallel clones, but only _1 batch_ will run at a time. A batch is defined as the contents of a `wit-manifest.json`.

This isn't the most parallel it could possibly be, but was practical to implement given my current understanding of the codebase, and should speed things up a little for everyone without any extra configuration.

Notes for the times I recorded:
- home network connection with no other jobs running on machine
- the initial (-a) package causes a lot of dependencies
- reference and partial clone use `wit init`
- reference requires WIT_WORKSPACE_REFERENCE to point at an earlier workspace
- partial parallel needs no extra configuration
- restore uses `wit restore`
- restore requires another workspace to point at to resolve the wit-lock.json
- the lock file used for restore is the same as generated by the other wit init runs

| [--reference](https://github.com/sifive/wit/pull/218)  | [partial-parallel](https://github.com/sifive/wit/pull/222) | [restore](https://github.com/sifive/wit/pull/211/) | Time |
| ------------- | ------------- | ------------- | ------------- |
| ✕  | ✕  | ✕ | 7m07s |
| ✅  | ✕  | ✕ | 4m13s |
| ✕  | ✅  | ✕ | 3m41s |
| ✅  | ✅  | ✕ | 1m09s|
| ✕  | ✕  | ✅ | 2m59s|
| ✅  | ✕  | ✅ | 0m13s|

### DIFF
--- a/lib/wit/dependency.py
+++ b/lib/wit/dependency.py
@@ -1,4 +1,5 @@
 import re
+import multiprocessing.dummy
 from datetime import datetime
 from pathlib import Path
 from typing import List  # noqa: F401
@@ -36,22 +37,18 @@ class Dependency:
         self.dependents = []  # type: List[Package]
         self.message = message
 
-    def resolve_deps(self, wsroot, repo_paths, download, source_map, packages, queue, errors):
+    def resolve_deps(self, wsroot, repo_paths, download, source_map, packages, queue, jobs):
         source_map = source_map.copy()
         packages = packages.copy()
         queue = queue.copy()
-        errors = errors.copy()
         subdeps = self.package.get_dependencies()
         log.debug("Dependencies for [{}]: [{}]".format(self.name, subdeps))
+
+        errors = self._parallel_clone(subdeps, wsroot, repo_paths, download, jobs)
+
         for subdep in subdeps:
-            try:
-                subdep.load(packages, repo_paths, wsroot, download)
-            except BadSource as e:
-                errors.append(e)
-                continue
-
+            subdep.load(packages, repo_paths, wsroot, download=False)
             sources_conflict_check(subdep, source_map)
-
             source_map[subdep.name] = subdep.package.resolve_source(subdep.source)
 
             if subdep.package.repo is None:
@@ -67,6 +64,21 @@ class Dependency:
         queue.sort(key=lambda tup: tup[0])
 
         return source_map, packages, queue, errors
+
+    def _parallel_clone(self, deps, wsroot, repo_paths, download, jobs):
+        errors = []
+
+        def do(dep):
+            try:
+                p = Package(dep.name, repo_paths)
+                p.load(wsroot, download, dep.source, dep.specified_revision)
+            except BadSource as e:
+                errors.append(e)
+
+        with multiprocessing.dummy.Pool(jobs) as pool:
+            pool.map(do, deps)
+
+        return errors
 
     def __key(self):
         return (self.source, self.specified_revision, self.name)
--- a/lib/wit/gitrepo.py
+++ b/lib/wit/gitrepo.py
@@ -78,7 +78,6 @@ class GitRepo:
     def clone(self, source, name):
         assert not GitRepo.is_git_repo(self.path), \
             "Trying to clone and checkout into existing git repo!"
-        log.info('Cloning {}...'.format(self.name))
 
         cmd = ["clone", *self._git_reference_options(), "--no-checkout", source, str(self.path)]
         proc = self._git_command(*cmd, working_dir=str(self.path.parent))
@@ -89,6 +88,7 @@ class GitRepo:
                 raise BadSource(name, source)
             else:
                 raise
+        log.info('Cloned {}'.format(self.name))
 
     def _git_reference_options(self):
         """
--- a/lib/wit/main.py
+++ b/lib/wit/main.py
@@ -68,7 +68,7 @@ def main() -> None:
             # These commands assume the workspace already exists. Error out if the
             # workspace cannot be found.
             try:
-                ws = WorkSpace.find(Path.cwd(), parse_repo_path(args))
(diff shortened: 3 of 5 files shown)


lib/wit/dependency.py
```diff
@@ -43,15 +44,12 @@ def resolve_deps(self, wsroot, repo_paths, download, source_map, packages, queue
...
-                subdep.load(packages, repo_paths, wsroot, download)
-            except BadSource as e:
-                errors.append(e)
-                continue
 
-            sources_conflict_check(subdep, source_map)
+        self._parallel_clone(subdeps, wsroot, repo_paths, download, errors)
 
+        for subdep in subdeps:
+            subdep.load(packages, repo_paths, wsroot, False)
```
richardxia, line 51: For readability purposes, since you've replaced original argument that had the variable name `download` with just a constant `False`, can you pass this argument as a named parameter?

```suggestion
            subdep.load(packages, repo_paths, wsroot, download=False)
```

lib/wit/dependency.py
```diff
@@ -68,6 +66,23 @@ def resolve_deps(self, wsroot, repo_paths, download, source_map, packages, queue
 
         return source_map, packages, queue, errors
 
+    def _parallel_clone(self, deps, wsroot, repo_paths, download, errors):
+        def do(dep):
+            try:
+                p = Package(dep.name, repo_paths)
+                p.load(wsroot, download, dep.source, dep.specified_revision)
+            except BadSource as e:
+                errors.append(e)
```
richardxia, line 75: I feel like it's pretty unusual in Python to pass in a list as an argument and expect the function to mutably append to it as needed. Could we change this function such that we don't take in the `errors` list as an argument, but we instead build up an errors list and just return it at the end? On the caller side you can still mutate the array, but I think it's best if the caller is responsible for that.

lib/wit/dependency.py
```diff
@@ -68,6 +66,23 @@ def resolve_deps(self, wsroot, repo_paths, download, source_map, packages, queue
...
+
+        threads = list()
+        for dep in deps:
+            t = threading.Thread(target=do, args=(dep,))
```
richardxia, line 79: One thing that I'm a little afraid of is that we might trip some rate limiting or DoS detection by spawning many threads that all simultaneously try to connect to GitHub. Would it make sense to cap the number of concurrent threads, as well as introduce some command line argument that forces single-threaded execution?

I think the easiest way to do a cap on concurrent threads is to use the `multiprocessing.dummy` module, which is an implementation of the normal Python `multiprocessing` module except that [it is implemented with threads](https://docs.python.org/3/library/multiprocessing.html#module-multiprocessing.dummy), and make use of the [`Pool`](https://docs.python.org/3/library/multiprocessing.html#using-a-pool-of-workers) class. You can basically do something like:

```py
with Pool(10) as pool:
    pool.map(do, deps)
```

lib/wit/dependency.py
jackkoenig: I agree that's better, we do do a bit of mutation of arguments in the codebase, but doing it less is ideal

jackkoenig:
This is awesome, and nice measurements. I agree with Richard's worries about tripping DoS detection, there could be a `-p` style argument to set the number of threads.

lib/wit/dependency.py
mmjconolly: Github's ratelimiting (for API atleast, couldn't find _anything_ for ssh clone) specifies everything per hour per user, so my guess is that spending 7 minutes downloading 100 repos is the same as spending 1 minute, as it tends to specify 'number of events' rather than 'concurrent events' (and in this diff you don't even get max concurrency).
Happy to put a option in to limit it, I do feel we should leave the default as 'as parallel as you can' unless we can query how much of the ratelimit is remaining for the time period I feel like we're artificially limiting ourselves.

richardxia: Looks like there's no specific limit, but also a caveat of "don't overload our servers": https://github.community/t5/GitHub-API-Development-and/Git-clone-limits-using-git-commands-vs-the-API-what-are-they/td-p/33623

I think I'd request that we still put in a parameter for setting a limit, since I am worried that with no limit we could run into a situation in the future where we're under a deadline and suddenly start running into rate limiting issues, and we can't run CI jobs while frantically trying to ship out a patch release of `wit` that adds in that parameter.

I'm mostly OK with leaving the default as "as parallel as you can", since for a user running this interactively, that should be fine, and even if we run into issues, we can tell the user "just set -j10" or something. The parameter is also important for CI, where the amount of typing/user experience of the wit CLI is less important as it should be captured in a script anyway, so we can afford to have more options.

mmjconolly:
The largest number of dependencies I found in a wit-manifest.json was 48, so I've set the default max parallelism to 64 which should be 'ought to be enough for anyone'. They can use `-j` to set it to 1 or something large if they like

mmjconolly:
I've now run this on a large shared machine, which has much faster network, more cpu cores, unknown speed disk.
 | [--reference](https://github.com/sifive/wit/pull/218)  | [partial-parallel](https://github.com/sifive/wit/pull/222) | [restore](https://github.com/sifive/wit/pull/211/) | Time (my laptop) | Time (big machine)
| ------------- | ------------- | ------------- | ------------- | ------------- | 
| ✕  | ✕  | ✕ | 7m07s | 3m46s |
| ✅  | ✕  | ✕ | 4m13s | 3m05s |
| ✕  | ✅  | ✕ | 3m41s | 2m20s |
| ✅  | ✅  | ✕ | 1m09s| 1m37s |
| ✕  | ✕  | ✅ | 2m59s| 0m41s |
| ✅  | ✕  | ✅ | 0m13s| 0m19s |

PR #224, opened 2020-04-20
Use python docker image based on alpine

mmjconolly:
Previous docker ubuntu-based image was 631mb, this alpine-based image is 123mb.
Won't work for wit version 0.12.0 and earlier, as it does not contain Java.

### DIFF
--- a/docker/wit.Dockerfile
+++ b/docker/wit.Dockerfile
@@ -1,25 +1,13 @@
-FROM ubuntu:xenial AS build
+FROM python:3.8-alpine
 
-RUN apt-get update && apt-get install -y \
-    git \
-    make \
-    python3-pip
+# Since we have no python deps outside the standard library,
+# just copy in the python files we need
+COPY ./lib/wit/*.py /wit/lib/wit/
 
-COPY . /wit/
-WORKDIR /wit
-RUN make install install_dir=/opt/sifive/wit
+RUN apk add git --no-cache                    && \
+    echo exec python3 -m wit '$@' > /bin/wit  && \
+    chmod +x /bin/wit
 
-# Actual image build
-FROM ubuntu:xenial
+ENV PYTHONPATH=/wit/lib:$PYTHONPATH
 
-WORKDIR /root/
-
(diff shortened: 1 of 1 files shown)


docker/wit.Dockerfile
```diff
@@ -1,25 +1,13 @@
...
-    python3-pip
+# Since we have no python deps outside the standard library,
+# just copy in the python files we need
+COPY ./lib/wit/*.py /wit/lib/wit/
 
-COPY . /wit/
-WORKDIR /wit
-RUN make install install_dir=/opt/sifive/wit
+RUN apk add git --no-cache                    && \
+    echo exec python3 -m wit '$@' > /bin/wit  && \
```
richardxia, line 8: Is there a reason for creating a new `/bin/wit` executable instead of doing what the [Makefile does](https://github.com/sifive/wit/blob/816e8738abe3e4cb0d7ea7d0bc34888085a9d454/Makefile#L9) by copying over the [`wit` executable shell script](https://github.com/sifive/wit/blob/816e8738abe3e4cb0d7ea7d0bc34888085a9d454/wit) from the top level of this repo? Is it just because you don't want to install `bash` in this Docker image?

I'm wondering if we can just install this like a normal Python library with `pip install .`, which, if set up correctly, will install any executables to the appropriate location (usually somewhere like `/usr/local/bin`). Within a virtual environment, the `pip install` command should install into just that virtual environment, to a directory that gets added to your path when you activate the virtual environment (`${venv_dir}/bin`).

Reusing the normal Python package distribution mechanism allows us to minimize having too many divergent ways of installing this package.

BTW, if replacing this line with `pip install` does "just work", then I'm fine with merging this as is, since I don't want us to rabbit hole on this topic.

mmjconolly: We don't need anything that the makefiles `make install` does, and that make target would probably get removed once we start publishing to pypi.

I thought about using the pip install, but that seemed more complicated as we need to create the package in a build image, then copy that into dest image just to do a pip install.

If we start looking at adding more deps to wit can make this more complicated

PR #225, opened 2020-04-20
Set version string to 0.13.x

mmjconolly:
Set version to `0.13.x` in master branch.
A separate 0.13.0 branch will be made which sets `0.13.0` there.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 install_root := $(PREFIX)
-version := $(shell cat lib/wit/version.py | grep -o '[.0-9]*')
+version := $(shell cat lib/wit/version.py | grep __version__ | cut -d= -f2 | tr -d ' "')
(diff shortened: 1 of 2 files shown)


richardxia:
I think this might actually break `setuptools` (although I haven't tried running it). A letter is not legal at that position, and the only letters you can use within the normal version components are `a`, `b`, and `rc` standing for "alpha", "beta", and "release candidate", respectively. Even when those letters are allowed, a digit must precede and succeed them, so you'd still be limited to writing something like `0.13.0a0`. It looks like you can also do `0.13.dev0`, which is under the [developmental releases](https://www.python.org/dev/peps/pep-0440/#developmental-releases) section.

See https://www.python.org/dev/peps/pep-0440/ for the gory details.

mmjconolly:
Changed to 0.13.dev0, tested I could locally install

PR #230, opened 2020-04-21
Use pip install for docker and make install

mmjconolly:
New docker image size is 129mb instead of 123mb, but an extra 6mb should be ok


mmjconolly:
ah, need to fixup the regress test

PR #235, opened 2020-04-30
Update typo in README.md

rmac-sifive:


jackkoenig:
The default branch is `0.13.x` which is why it opened up against that but we make PRs against master

PR #236, opened 2020-05-06
Add /bin/bash to github actions docker container

mmjconolly:
Since we moved our actions docker container to alpine linux, we should use /bin/sh instead of bash

### DIFF
--- a/docker/wit.Dockerfile
+++ b/docker/wit.Dockerfile
@@ -3,7 +3,7 @@ FROM python:3.8-alpine
 COPY ./lib /wit/
 WORKDIR /wit
 
-RUN apk add git --no-cache && \
+RUN apk add git bash --no-cache && \
     python3 -m pip install .
 
 CMD /bin/sh


mmjconolly:
I haven't actually tested it yet

mmjconolly:
I kinda think life is easier if we require all images to have `/bin/bash`

richardxia:
> I kinda think life is easier if we require all images to have `/bin/bash`

Yeah, I think that is not a terrible thing to add. Alternatively, we could write this in Python given that we already depend on Python? But probably faster and fine to just install `bash` into the Docker image.

mmjconolly:
Adding /bin/bash instead

richardxia:
Sorry, I should have paid closer attention. You actually need to add `bash` to the [Dockerfile corresponding to the GitHub Action](https://github.com/sifive/wit/blob/81d586c25488433c0324d49fea2e4324c714531c/actions/wit/Dockerfile), not wit's own Dockerfile.

mmjconolly:
My hunch is that it's because we have `FROM sifive/wit:v0.13.0` in `actions/wit/Dockerfile`, where 0.13 is older than where we added bash

PR #239, opened 2020-05-13
bash to sh

davidmlw:
The docker image has no bash. Use sh instead.

### DIFF
--- a/actions/wit/entrypoint.sh
+++ b/actions/wit/entrypoint.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
 
 set -euxo pipefail
 


mmjconolly:
Please target branches at `master`

mmjconolly:
In `master` we handle this by adding bash to the actions container
https://github.com/sifive/wit/blob/9fdc590b4af08bb5094842df868fbc8aa9391597/actions/wit/Dockerfile#L4

richardxia:
By the way, this change also won't work correctly on all systems, since I used bash-specific syntax (the `[[ ]]`) which are not part of the POSIX standard. I know for a fact that `dash` installed as `/bin/sh` on Ubuntu-based systems will not support that syntax.

If you just use the wit action targeting the latest commit on master, you should get bash installed in the Docker container and it should work around this issue.

davidmlw:
> By the way, this change also won't work correctly on all systems, since I used bash-specific syntax (the `[[ ]]`) which are not part of the POSIX standard. I know for a fact that `dash` installed as `/bin/sh` on Ubuntu-based systems will not support that syntax.
> 
> If you just use the wit action targeting the latest commit on master, you should get bash installed in the Docker container and it should work around this issue.

I see. Seems better way is to apk add bash

PR #240, opened 2020-05-13
Highlight pip install in README.md

mmjconolly:
How do we feel about changing default branch to `master` after this readme change?

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,35 +17,18 @@ may exist only once.
 ## How does Wit deduplicate packages?
 When multiple versions of the same package are requested, Wit chooses the latest requested version, making sure the selected version's commit is a descendant of every other requested versions' commits.
 
-## Getting started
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -17,35 +17,18 @@ may exist only once.
 ## How does Wit deduplicate packages?
 When multiple versions of the same package are requested, Wit chooses the latest requested version, making sure the selected version's commit is a descendant of every other requested versions' commits.
 
-## Getting started
-
-The best way to learn wit is to check out the [tutorial](doc/tutorial.md).
-
 ### Installation
```
richardxia, line 20: You'll want to change this to be a level-2 heading instead of a level-3 heading, since the way it renders on GitHub now makes it look like it belongs under the "How does Wit deduplicate packages?" section.

PR #244, opened 2020-06-05
Consolidate json format parsing

mmjconolly:
Unlike https://github.com/sifive/wit/pull/242 this does not change the formats.

These 3 file types now use the same parse/unparse/write/read code
- wit-manifest.json
- wit-lock.json
- wit-manifest.json

Most of the change is the addition of `repo_entries.py` and the rest of the changes are to use the new code from that file.

While formats themselves remain unchanged, this refactoring should make it easier for us to support new formats while maintaining compatibility.

As a future direction, I'm thinking that by checking the path name of a file we can know to write it as a, for example, `wit-lock.json`-style. 
Where as if it's called `new-format.toml` perhaps we could read out the version number from the current on-disk file to see what version we should write out.

### DIFF
--- a/lib/wit/repo_entries.py
+++ b/lib/wit/repo_entries.py
@@ -1,0 +1,114 @@
+import json
+import sys
+from enum import Enum
+from pathlib import Path
+from typing import List
+
+
+# The intent of Format, RepoEntry and List[RepoEntry] is that no other
+# parts of the codebase knows that json is used as the on-disk format, or know
+# any of the field names.
+
+# Version numbers will be encoded in the format from '3' onwards.
+class Format(Enum):
+    Lock = 1
+    Manifest = 2
+
+    @staticmethod
+    def from_path(path: Path):
+        if path.name == "wit-lock.json":
+            return Format.Lock
+        if path.name == "wit-workspace.json" or path.name == "wit-manifest.json":
+            return Format.Manifest
+        raise Exception("Unknown format for {}".format(str(path)))
+
+
+class RepoEntry:
+    def __init__(self, checkout_path, revision, remote_url, message=None):
+        # The path to checkout at within the workspace.
+        # JSON field name is 'name'.
+        self.checkout_path = checkout_path
+
+        # Desired revision that exists in the history of the below remote.
+        # JSON field name is 'commit'
+        self.revision = revision
+
+        # Url (or local fs path) for git to clone/fetch/push.
+        # JSON field name is 'source'
+        self.remote_url = remote_url
+
+        # A comment to leave in any serialized artifacts.
+        # Optional. JSON field name is '//'
+        self.message = message
+
+    def __repr__(self):
+        return str(self.__dict__)
+
+
+# OriginalEntry encodes the RepoEntry for both Lock and Manifest formats.
+class OriginalEntry():
+    @staticmethod
+    def to_dict(entry: RepoEntry) -> dict:
+        d = {
+            "name": entry.checkout_path,
+            "commit": entry.revision,
+            "source": entry.remote_url,
+        }
+        if entry.message:
+            d["//"] = entry.message
+        return d
+
+    @staticmethod
+    def from_dict(data: dict) -> RepoEntry:
+        return RepoEntry(data["name"],
+                         data["commit"],
+                         data.get("source"),  # 'repo path' cli option needs this optional
+                         data.get("//"))  # optional
+
+
+# Utilities for List[RepoEntry]
+class RepoEntries:
+    @staticmethod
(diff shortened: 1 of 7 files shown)


lib/wit/repo_entries.py
```diff
@@ -0,0 +1,116 @@
...
+            print("Failed to parse json in {}:{}: {}".format(path, rev, e.msg))
+            sys.exit(1)
+
+        entries = []
+        fmt = Format.from_path(path)
+        for element in fromtext:
+            if fmt is Format.Manifest:
+                entry = element
+            if fmt is Format.Lock:
+                entry = fromtext[element]
```
richardxia, line 106: I realized that `fromtext` is either a list or a dict at the top level depending on which type of wit JSON file it is and that `for` loop iteration itself has different behavior depending on the file format, since with lists, you iterate over the list items, but with dicts, you iterate over the dictionary keys. This is kind of a surprising thing to read, even for someone who is aware of the different multiple file formats, and I was a bit confused about what was going on in the loop body because `fromtext` and `element` are different types depending on the input file.

Could I suggest that we hoist the `if fmt is Format.*` checks outside of the for loop so that it's clearer, and that you could iterate explicitly on the dictionary values for the dict case?

E.g.
```python
if fmt is Format.Manifest:
    for entry in fromtext:
        entries.append(OriginalEntry.from_dict(entry))
elif fmt is Format.Lock:
    for entry in fromtext.values():
        entries.append(OriginalEntry.from_dict(entry))
```

lib/wit/repo_entries.py
```diff
@@ -0,0 +1,116 @@
...
+
+# The intent of Format, RepoEntry and List[RepoEntry] is that no other
+# parts of the codebase knows that json is used as the on-disk format, or know
+# any of the field names.
+
+# Version numbers will be encoded in the format from '3' onwards.
+class Format(Enum):
+    unknown = 0
+    Lock = 1
+    Manifest = 2
```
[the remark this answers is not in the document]
mmjconolly: It doesn't have to exist, it's something I see in other languages usage of enums and still haven't learned many python-isms

mmjconolly:
> It especially feels that most of the functionality of Manifest, Lock, and maybe even Package? is redundant with RepoEntry/RepoEntries. I think one difference between some of the data types is that some just represent the JSON files (the "spec"), and other represent internal state that changes during the course of the Wit operation?

The `dependency/package -> repoentry` and `workspace/manifest -> [repoentry]` is something I've been thinking about for awhile and feel like it could be cleaned up. My main intent with this PR was to change as little as possible and follow up with more refactors over time

lib/wit/repo_entries.py
mmjconolly: yep will do

lib/wit/repo_entries.py
```diff
@@ -0,0 +1,116 @@
...
+class RepoEntries:
+    @staticmethod
+    def write(path: Path, entries: List[RepoEntry]):
+        fmt = Format.from_path(path)
+        if fmt is Format.Manifest:
+            data = [OriginalEntry.to_dict(e) for e in entries]
+            json_data = json.dumps(data, sort_keys=True, indent=4) + "\n"
+        if fmt is Format.Lock:
+            ordered_data = OrderedDict((e.checkout_path, OriginalEntry.to_dict(e)) for e in entries)
+            json_data = json.dumps(ordered_data, sort_keys=True, indent=4) + "\n"
```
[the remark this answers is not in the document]
mmjconolly: yep I think I just applied too much copy/paste from how it used to work

richardxia:
> The `dependency/package -> repoentry` and `workspace/manifest -> [repoentry]` is something I've been thinking about for awhile and feel like it could be cleaned up. My main intent with this PR was to change as little as possible and follow up with more refactors over time

Yeah, I understand, and I didn't mean to imply that we should change it now, but I thought that even a tiny bit of documentation could help provide clarity on where we're going. I guess don't worry about it, especially if there are some other PRs coming up that are going to change them anyway.

lib/wit/repo_entries.py
```diff
@@ -0,0 +1,115 @@
...
+            print("Failed to parse json in {}:{}: {}".format(path, rev, e.msg))
+            sys.exit(1)
+
+        entries = []
+        fmt = Format.from_path(path)
+        if fmt is Format.Manifest:
+            for entry in fromtext:
+                entries.append(OriginalEntry.from_dict(entry))
+        if fmt is Format.Lock:
+            print(fromtext)
```
richardxia, line 104: Leftover debug print statement?

richardxia:
Whoops, I had a stale browser tab open where it looked like I had not submitted an updated review.

Issue #248, opened 2020-07-24
wit 'foreach' required quotes around some commands


mmjconolly:
`wit foreach 'git stash'` should also work as `wit foreach git stash`

PR #249, opened 2020-07-24
wit foreach shouldn't require quotes for 'cmd'

mmjconolly:
Fixes https://github.com/sifive/wit/issues/248

### DIFF
--- a/lib/wit/main.py
+++ b/lib/wit/main.py
@@ -124,12 +124,11 @@ def foreach(ws, args):
         log.info("Entering '{}'".format(pkg.name))
 
         location = str(ws.root / pkg.name)
-        proc = subprocess.run([args.cmd, *args.args], shell=True,
-                              env=env, cwd=location, universal_newlines=True)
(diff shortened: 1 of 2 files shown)


lib/wit/main.py
```diff
@@ -123,13 +123,13 @@ def foreach(ws, args):
 
         log.info("Entering '{}'".format(pkg.name))
 
+        command = " ".join([args.cmd] + args.args)
         location = str(ws.root / pkg.name)
-        proc = subprocess.run([args.cmd, *args.args], shell=True,
+        proc = subprocess.run(command, shell=True,
                               env=env, cwd=location, universal_newlines=True)
```
jackkoenig, line 129: Does this work when the foreached command itself has quotes? eg.

```
$ wit foreach git grep 'Hello World'
```

richardxia: Might be good to add a test case for this.

mmjconolly: good catch

PR #250, opened 2020-08-08
actions/init: add a new rule capable of initializing a wit workspace

terpstra:
This makes checking out a wit workspace a one-liner:
```yaml
  - uses: sifive/wit/actions/init@v0.13.2
```

If you want to run wake on the workspace, you can also specify an environment package like this:
```yaml
 jobs:
   test:
     name: Scala compile
     runs-on: ubuntu-latest

     steps:
     - name: Wit Init
       uses: sifive/wit/actions/init@v0.13.2
       with:
         additional_packages: git@github.com:sifive/environment-blockci-sifive.git::0.7.0

     - name: Run wake scala compile
       uses: sifive/environment-blockci-sifive/actions/wake@0.7.0
       with:
         command: -x 'compileScalaModule myScalaModule | getPathResult'
 ```

### DIFF
--- a/actions/init/README.md
+++ b/actions/init/README.md
@@ -9,6 +9,20 @@ See [action.yml](./action.yml) for a detailed list of input parameters.
 
 ## Example Usage
 
+In the simplest scenario, something like this would work:
+```yaml
+jobs:
+  test:
+    name: Print daffy in workspace
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: sifive/wit/actions/init@v0.13.2
+    - run: echo "Daffy duck"
+```
+
+However, if you are using wake and have an environment packge to include,
+this is probably a more representative example:
 ```yaml
 jobs:
   test:
--- a/actions/init/README.md
+++ b/actions/init/README.md
@@ -1,9 +1,9 @@
 # Wit Init Action
 
 This [GitHub Action](https://github.com/features/actions) will initialize a
-wit workspace for a repository with a wit-manifest.json. An environment
-package should be specified to describe the build environment. By default,
-environment-example-sifive is used to support reading a JSON description.
+wit workspace for a repository with a wit-manifest.json.  When used in a
+complex project, an environment package should probably be specified to
+describe the build environment using the additional_packages argument.
 
 See [action.yml](./action.yml) for a detailed list of input parameters.
 
@@ -19,7 +19,7 @@ jobs:
     - name: Wit Init
       uses: sifive/wit/actions/init@v0.13.2
       with:
-        environment: git@github.com:sifive/environment-blockci-sifive.git::0.7.0
+        additional_packages: git@github.com:sifive/environment-blockci-sifive.git::0.7.0
 
     - name: Run wake scala compile
       uses: sifive/environment-blockci-sifive/actions/wake@0.7.0
--- a/actions/init/README.md
+++ b/actions/init/README.md
@@ -1,0 +1,28 @@
+# Wit Init Action
+
+This [GitHub Action](https://github.com/features/actions) will initialize a
+wit workspace for a repository with a wit-manifest.json. An environment
+package should be specified to describe the build environment. By default,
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by terpstra** [comment]
Once this is merged, I would greatly appreciate a new tag (v0.13.2) be assigned to the merge commit.

File: actions/init/README.md [hunk 1]
```diff
@@ -0,0 +1,28 @@
...
+    steps:
+    - name: Wit Init
+      uses: sifive/wit/actions/init@v0.13.2
+      with:
+        environment: git@github.com:sifive/environment-blockci-sifive.git::0.7.0
+
+    - name: Run wake scala compile
+      uses: sifive/environment-blockci-sifive/actions/wake@0.7.0
+      with:
+        command: -x 'compileScalaModule myScalaModule | getPathResult'
```
jackkoenig, line 27: While I appreciate the more complete example, is this the place to have an action from another repo?

actions/init/action.yml
```diff
@@ -0,0 +1,57 @@
...
+      The location where wit will initialize the workspace. All repositories,
+      including the invoking repository, will be subdirections of workspace.
+    required: false
+    default: '.'
+  environment:
+    description: |
+      The environment package to use when building this repository. This project
+      will be automatically added to the wit workspace
+    required: false
+    default: 'git@github.com:sifive/environment-example-sifive.git'
```
jackkoenig, line 15: 1. Should this be called `environment` when it can work for any packages? I grant `environment` may better describe the pattern, but in theory you could have any number of packages here.

2. I don't think there should be a default

terpstra:
I've created a different PR for the v0.13.x branch, to take the tag off of.

actions/init/README.md
terpstra: I think anyone who uses this rule is going to want to use the two of them together. This comment is copy-pastable, to encourage people to do the right thing.

actions/init/action.yml
terpstra: Sure, I can remove the default.

If I could figure out how to accept a github action multiple times; ie:
```yml
- uses: sifive/wit/actions/init@v0.13.2
       with:
         extra: pkg1
         extra: pkg2
```

that would obviously be better. However, it does not appear that is possible yet. When/if they add that (seems github is actively improving this stuff) we should add another more general purpose option.

For now, for the use-cases we are actually catering towards, I think environment is a good choice.

actions/init/README.md
mmjconolly: This could be something generic,
```
- name: "run a command in the workspace"
  with:
    command: cowsay hello
```

terpstra: You say that wake has nothing to do with wit. In theory, that's true. In practice, if you are using wit, you are using wake. Someone who comes across this PR and/or docs is going to want to do both things.

actions/init/README.md
```diff
@@ -0,0 +1,28 @@
...
+jobs:
+  test:
+    name: Scala compile
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Wit Init
+      uses: sifive/wit/actions/init@v0.13.2
+      with:
+        environment: git@github.com:sifive/environment-blockci-sifive.git::0.7.0
```
mmjconolly, line 22: Can this be `additionalPackages`, wit doesn't know what an environment is

terpstra: Fine. Let me see if I can figure how to get it to work with more than one.

actions/init/action.yml
terpstra: So the above syntax (repeated 'extra') is illegal and causes the yml file to be ignored. I've changed it to a space-separated additional_packages argument. It appears to work.

actions/init/README.md
terpstra: I've changed it to a space-separated additional_packages argument.

actions/init/README.md
terpstra: I've changed the documentation to include both examples.

PR #251, opened 2020-08-10
actions/init: add a new rule capable of initializing a wit workspace

terpstra:
This is a cherry-pick of #250 to the 0.13.x branch.

This makes checking out a wit workspace a one-liner:
```yaml
  - uses: sifive/wit/actions/init@v0.13.2
```

If you want to run wake on the workspace, you can also specify an environment package like this:
```yaml
 jobs:
   test:
     name: Scala compile
     runs-on: ubuntu-latest

     steps:
     - name: Wit Init
       uses: sifive/wit/actions/init@v0.13.2
       with:
         additional_packages: git@github.com:sifive/environment-blockci-sifive.git::0.7.0

     - name: Run wake scala compile
       uses: sifive/environment-blockci-sifive/actions/wake@0.7.0
       with:
         command: -x 'compileScalaModule myScalaModule | getPathResult'
 ```


terpstra:
Note: if you try to this action as in the example, it will not work until the v0.13.2 tag has been applied.
Until then, you can try it using a raw hash like here: https://github.com/sifive/sifive-blocks/pull/158/files
... but that sifive-blocks PR can obviously not be merged until the hashes are replace by v0.13.2...

PR #253, opened 2020-08-12
Error out gracefully when failing to clone a dependency

mmjconolly:
This addresses the case where a package's dependency is unable to be cloned, but where it would still attempt to checkout a revision.
The message looked like
```
File "/usr/local/lib/python3.8/site-packages/wit/package.py", line 133, in checkout
    current_origin = self.repo.get_remote()
AttributeError: 'NoneType' object has no attribute 'get_remote'
```

This now returns errors after the parallel clone more eagerly
Error now looks like:
```
Creating new workspace [/home/user/workspace]
Cloned product1
The workspace now depends on 'product1::abcd123'
Cloned foo

--- ERROR ---
Bad remote for 'baa':
  git@github.com:sifive/baa
```

### DIFF
--- a/lib/wit/dependency.py
+++ b/lib/wit/dependency.py
@@ -4,7 +4,6 @@ from datetime import datetime
 from pathlib import Path
 from typing import List  # noqa: F401
 from .common import WitUserError
-from .gitrepo import BadSource
 from .package import Package
 from .repo_entries import RepoEntry
 from .witlogger import getLogger
--- a/lib/wit/dependency.py
+++ b/lib/wit/dependency.py
@@ -45,6 +45,8 @@ class Dependency:
         log.debug("Dependencies for [{}]: [{}]".format(self.name, subdeps))
 
         errors = self._parallel_clone(subdeps, wsroot, repo_paths, download, jobs)
+        if len(errors) > 0:
+            return {}, [], [], errors
 
         for subdep in subdeps:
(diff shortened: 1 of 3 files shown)


mmjconolly:
> LGTM!
> 
> I wish there were a clean way for us to augment the underlying Git message, since it's really not clear that `Bad remote` sometimes means "Permission denied", but I'm not sure if it'd be a good idea to try to match on the strings Git prints out. I suppose that something like this would be cleaner with a lower-level API like libgit2?

I think github actually 404s for a permission denied, so that they don't leak the existence of things

richardxia:
> I think github actually 404s for a permission denied, so that they don't leak the existence of things

Right, but even if that's the case, we could offer a suggestion or two to the user on how to address it, e.g. "Do you have permissions to this repo, or did you misspell the repo?". My main concern with string matching on "Bad remote" is that it could be dependent on the version of underneath or even locale settings, if Git supports that.

mmjconolly:
> > I think github actually 404s for a permission denied, so that they don't leak the existence of things
> 
> Right, but even if that's the case, we could offer a suggestion or two to the user on how to address it, e.g. "Do you have permissions to this repo, or did you misspell the repo?". My main concern with string matching on "Bad remote" is that it could be dependent on the version of underneath or even locale settings, if Git supports that.

Ah I see what you mean, yeah libgit2 would help with that

PR #256, opened 2020-09-02
Use git submodules as a source of dependencies

mmjconolly:
## what
* If repository lacks `wit-manifest.json`, use git submodules to collect dependency data
* `wit-manifest.json` is used in preference to git submodules, for repositories that have both
* `wit add-dep`/`wit update-dep` on a repository that only has git submodules errors out with a message, user would use regular git submodules commands to update those dependencies. These are the only two commands that should be impacted.

## why
* in repositories that already keep a wit-manifest.json and a list of submodules manually sync-ed, we could delete the wit-manifest.json file (presuming they used a new enough wit version). It might not make sense in all cases to do this but the option is there

## how
When wit is performing a `resolve` to determine which is the correct commit to checkout, it checks for the wit-manifest as usual, if this is missing it checks for a `.gitmodules` file.
The .gitmodules file provides the list of remote url and the within-repository path, for each submodule.
The within-repository path is used to ask the git index what the pointed-to commit is for the submodule.

## relationship to submodules
This could be viewed as creating a read-only on-demand wit-manifest.json from the submodule information.
Wit makes no attempt to write submodule information. 
Submodules do appear in the wit-lock.json, but wit does not know they are submodules.

### DIFF
--- a/lib/wit/gitrepo.py
+++ b/lib/wit/gitrepo.py
@@ -4,7 +4,10 @@ import subprocess
 from pathlib import Path
 from pprint import pformat
 import re
+import os
+import sys
 from .common import WitUserError
+from collections import OrderedDict
 from .witlogger import getLogger
 from typing import List, Set  # noqa: F401
 from functools import lru_cache
@@ -46,10 +49,12 @@ class GitRepo:
     Note there can be multiple GitRepo objects for the same package
     """
     PKG_DEPENDENCY_FILE = "wit-manifest.json"
+    SUBMODULE_FILE = ".gitmodules"
 
     def __init__(self, name, wsroot: Path):
         self.name = name
         self.path = wsroot / name
+        self.wsroot = wsroot
         # Cache known hashes for quick lookup
         self._known_hashes = set()  # type: Set[str]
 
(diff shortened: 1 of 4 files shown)


lib/wit/gitrepo.py
```diff
@@ -242,14 +246,91 @@ def is_ancestor(self, ancestor, current=None):
...
-            log.debug("No dependency file found in repo [{}:{}]".format(revision,
+            log.debug("No wit dependency file found in repo [{}:{}]".format(revision,
                       self.path))
             return []
         return RepoEntries.parse(proc.stdout, Path(GitRepo.PKG_DEPENDENCY_FILE), revision)
 
+    def _read_submodules_from_commit(self, revision) -> List[RepoEntry]:
+        proc = self._git_command("show", "{}:{}".format(revision, GitRepo.SUBMODULE_FILE))
+        if proc.returncode:
+            log.debug("No .gitmodule file found in repo [{}:{}]".format(revision, self.path))
```
richardxia, line 265:
```suggestion
            log.debug("No .gitmodules file found in repo [{}:{}]".format(revision, self.path))
```

lib/wit/gitrepo.py
```diff
@@ -242,14 +246,91 @@ def is_ancestor(self, ancestor, current=None):
...
+            return []
+
+        log.debug("{}:{} does not have wit-manifest.json, "
+                  "reading dependencies from git submodules instead"
```
richardxia, line 269:
```suggestion
                  "reading dependencies from .gitmodules instead"
```

Nit: I think explicitly naming the file you will read from is probably clearer than saying "git submodules", especially given that the debug-level log messages are probably only meant for wit maintainers to use.

lib/wit/gitrepo.py
```diff
@@ -242,14 +246,91 @@ def is_ancestor(self, ancestor, current=None):
...
+                  .format(self.name, revision))
+
+        # Use the 'git config' parser to read the submodule contents.
+        # Output is of the form:
+        #     submodule.$NAME.path $PATH
+        #     submodule.$NAME.url  $REMOTE
+        proc = self._git_command("config", "-f-", "--get-regex", "submodule..*",
```
richardxia, line 276:
```suggestion
        proc = self._git_command("config", "-f-", "--get-regex", r"submodule\..*",
```

For safety, we may want to explicitly escape the `.` to treat it as a literal `.`.

lib/wit/gitrepo.py
```diff
@@ -242,14 +246,91 @@ def is_ancestor(self, ancestor, current=None):
...
+                                 input=proc.stdout)
+        self._git_check(proc)
+
+        paths_by_name = {}
+        urls_by_name = {}
+
+        path_r = re.compile('^submodule.(.*).path (.*)$')
```
richardxia, line 283:
```suggestion
        path_r = re.compile(r'^submodule\.(.*)\.path (.*)$')
```

Same here with the regex.

lib/wit/gitrepo.py
```diff
@@ -242,14 +246,91 @@ def is_ancestor(self, ancestor, current=None):
...
+        for line in proc.stdout.splitlines():
+            m = path_r.match(line)
+            if m:
+                paths_by_name[m.group(1)] = m.group(2)
+
+        url_r = re.compile('^submodule.(.*).url (.*)$')
```
richardxia, line 289:
```suggestion
        url_r = re.compile(r'^submodule\.(.*)\.url (.*)$')
```

lib/wit/gitrepo.py
richardxia: Also, thinking about this regex a bit more, is there a way Git itself escapes `.` in a submodule's name? I know that it's definitely legal to include a `.` in a Git repo's name. If Git does escaping of `.`, then we'll need to unescape it when we build the name -> path/url mappings.

lib/wit/gitrepo.py
```diff
@@ -242,14 +246,91 @@ def is_ancestor(self, ancestor, current=None):
...
+            if m:
+                urls_by_name[m.group(1)] = m.group(2)
+
+        if len(paths_by_name) != len(urls_by_name):
+            log.error("Error matching paths with urls in {}/{}"
+                      .format(self.name, GitRepo.SUBMODULE_FILE))
+            sys.exit(1)
+
+        submodules = []
+        for name_key, path in paths_by_name.items():
```
richardxia, line 301: I'm a little bit concerned by the ordering of iteration here, since that could cause nondeterministic results. Since Python 3.5 dictionaries are unordered, could you explicitly use `OrderedDict` for these?

lib/wit/gitrepo.py
```diff
@@ -242,14 +246,91 @@ def is_ancestor(self, ancestor, current=None):
...
+            submodules.append(RepoEntry(name, submodule_ref, url))
+
+        return submodules
+
+    def _get_submodule_pointer(self, revision, path):
+        """
+        Get the submodule pointer commit in the index.
+        Note: This is NOT always the currently checked-out commit for a submodule.
+        The 'git ls-tree' output is defined as:
+           ^<mode> space <type> space <object> tab <file>$
```
richardxia, line 326:
```suggestion
           ^<mode> space <type> space <git object hash> tab <file>$
```

I think explicitly calling this the "git object hash" will make this a bit more understandable by people who are less familiar with deeper git internals.

lib/wit/gitrepo.py
```diff
@@ -242,14 +246,91 @@ def is_ancestor(self, ancestor, current=None):
...
+            url = urls_by_name[name_key]
+            name = name_key
+            if "/" in name:
+                # Wit only supports a 'flat' checkout pattern.
+                # By default, git submodules uses the relative checkout path to name the submodule.
+                # The user can add an explict name, but often do not.
+                # So if the submodule happens to have a name that looks like a nested path,
+                # use the the final path component in the remote url instead as the path-based names
+                # as sometimes they're easy-to-clash names like docs/html.
+                name = os.path.basename(url).replace(".git", "")
```
richardxia, line 315: For safety, how about we use

```suggestion
                name = re.sub(r"\.git$", "", os.path.basename(url))
```

richardxia (commented):
This looks pretty good to me!

On a more practical note, do you know if this will cause us to add additional Git repos to the wit workspace for any of our larger, internal repositories? I'm just wondering if there are any repositories which do not have a `wit-manifest.json` file but have Git submodules that we may not want to pull into the workspace. I'd like to avoid cases like in rocket-chip where the full Git submodule dependency tree pulls in the entirety of the full GCC Git repository, which is a pretty heavy thing to pull in. rocket-chip specifically will not have an issue because it has a `wit-manifest.json` file, but I want to make sure we don't cause any other large repos like that to get pull in like this.

mmjconolly:
> do you know if this will cause us to add additional Git repos to the wit workspace for any of our larger, internal repositories? I'm just wondering if there are any repositories which do not have a `wit-manifest.json` file but have Git submodules that we may not want to pull into the workspace.

In the current scenario, where all the repositories that we care about have wit-manifest.json, as far as I've noticed we only add a couple of documentation repositories (like freedom-metal/docs)

> I'd like to avoid cases like in rocket-chip where the full Git submodule dependency tree pulls in the entirety of the full GCC Git repository, which is a pretty heavy thing to pull in. rocket-chip specifically will not have an issue because it has a `wit-manifest.json` file, but I want to make sure we don't cause any other large repos like that to get pull in like this.

I guess we can look at that if (when) we decide to delete the wit-manifest on rocket-chip.
(I just did a recursive submodule update on rocket-chip, couldn't see a gcc?)

lib/wit/gitrepo.py
mmjconolly: It allows the dot in the submodule name, but doesn't escape it

richardxia:
> I guess we can look at that if (when) we decide to delete the wit-manifest on rocket-chip.
> (I just did a recursive submodule update on rocket-chip, couldn't see a gcc?)

We must have removed that at some point, probably after GCC upstream finally accepted the RISC-V patches. That at least simplifies this problem a lot!

richardxia (approved):
LGTM!

Issue #234, opened 2020-04-28
WIT_WORKSPACE_REFERENCE only works with absolute paths


erikdanie:
I tried doing a wit init with WIT_WORKSPACE_REFERENCE pointing directly to a workspace and it failed with 
`stderr: [Cloning into '/path/to/new-wkspc/.wit/repo'...
fatal: Could not switch to 'old-wkspc/': No such file or directory]`

PR #258, opened 2020-09-14
Disallow relative paths in WIT_WORKSPACE_REFERENCE

mmjconolly:
Resolve a relative path in `$WIT_WORKSPACE_REFERENCE` into an absolute path.

Also, mention that the variable exists in `--help` as the markdown docs are less convenient 

For https://github.com/sifive/wit/issues/234

### DIFF
--- a/lib/wit/main.py
+++ b/lib/wit/main.py
@@ -21,6 +21,7 @@ from .inspect import inspect_tree
 from pathlib import Path
 from typing import cast, List, Tuple  # noqa: F401
 from .common import error, WitUserError, print_errors
+from .env import git_reference_workspace
 from .gitrepo import GitRepo, GitCommitNotFound
 from .manifest import Manifest
 from .package import WitBug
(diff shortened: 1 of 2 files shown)


richardxia (approved):
This is relative to the working directory? I guess this LGTM, although it feels a little bit fragile for a user to set this to a relative path, but I guess I see no harm in allowing the user to do that.

Note that in Python 3.5, [`Path.resolve()`](https://docs.python.org/3.5/library/pathlib.html#pathlib.Path.resolve) method will throw an exception if the path does not actually exist on the system. Also note that the `strict` option to that method was only added in 3.6 or higher. Frustratingly enough, when they added the `strict` option in Python 3.6, they also apparently changed the default behavior to be _not strict_.

I am seeing [`Path.cwd() / your_path` as a suggested workaround](https://stackoverflow.com/a/44569249) that is compatible with both Python 3.5 and 3.6+, although this alternative does not resolve symlinks.

mmjconolly:
How about if I change it to error-out when using relative absolute paths? That removes any doubt

richardxia:
Sounds good to me!

PR #260, opened 2020-09-22
Allow submodules to be ignored by reading .gitmodule metadata

mmjconolly:
Allow a user to specify that a git submodule of a repository should not be considered as input to the dependency resolution algo, by adding a key/val pair to `.gitmodules`

```
$ cat workspace/foo/.gitmodules
[submodule "baa"]
    path = baa
    url = https://github.com/baa-corp/baa.git
    wit = ignore
```

### DIFF
--- a/doc/how-to-guides.adoc
+++ b/doc/how-to-guides.adoc
@@ -110,3 +110,39 @@ restore
     $ ls
     wit-lock.json wit-workspace.json
     $ wit restore
+
+== Specify dependencies via git submodules
+
+Some repositories that you would like to use as a dependency already use git submodules.
+Wit can read the git subomodule metadata from the repository as if it were a `wit-manifest.json`
+file.
+Wit will only read submodule data when there is no `wit-manifest.json` file.
+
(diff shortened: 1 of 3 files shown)


jackkoenig (commented):
I like the docs and clever idea for ignoring, I didn't even considering trying to add metadata to the `.gitmodules` file. Do you know for sure that it's safe?

mmjconolly:
> I like the docs and clever idea for ignoring, I didn't even considering trying to add metadata to the `.gitmodules` file. Do you know for sure that it's safe?

Reasonably confident that it's safe, the .gitmodules file is under the hood a [git config](https://git-scm.com/docs/git-config) format file (ie, section/key/val) that happens to be used for submodules.

I did some manual tests that preserve/ignore the extra metadata:
* added a new submodule after a previous one already had 'ignore'
* update a submodule that has 'ignore'
* git submodule status
* git submodule foreach

richardxia (approved):
LGTM, and nice docs!

PR #261, opened 2020-10-28
Wit 0.14.1

mmjconolly:
Pull in latest from master, set version and tag

This will allow the use of `wit = ignore` in .gitmodules


richardxia (approved):
LGTM!

